PR #16, opened 2020-05-05
more microphysics parameters

trontrytel:
Because there is never enough of microphysics parameters.

@charleskawczynski - I wasn't sure if there are other places I need to touch in order to add those parameters.  I'm still working on the tests using them in ClimateMachine repo.

### DIFF
--- a/src/Atmos.jl
+++ b/src/Atmos.jl
@@ -52,12 +52,29 @@ end # module SubgridScale
 
 module Microphysics
 
-export MP_n_0,
+export n0,
+       μ_sno,
+       ν_sno,
+       r0,
+       m0,
+       me,
+       χm,
+       Δm,
+       a0,
+       ae,
+       χa,
+       Δa,
+       v0,
+       ve,
+       χv,
+       Δv,
        C_drag,
        τ_cond_evap,
+       τ_sub_dep,
        q_liq_threshold,
        τ_acnv,
-       E_col,
+       E,
+       r_ice_snow,
        a_vent,
        b_vent,
        K_therm,
@@ -65,8 +82,53 @@ export MP_n_0,
        ν_air,
        N_Sc
 
-"""Marshall-Palmer distribution n_0 coeff [1/m4]"""
-function MP_n_0 end
+"""Marshall-Palmer distribution n_0 coeff for rain or snow [1/m4]"""
+function n0 end
+
+"""coefficient to compute Marshall-Palmer distribution coefficient n_0(ρq_snow/ρ0) for snow [1/m4]"""
+function μ_sno end
+
+"""coefficient to compute Marshall-Palmer distribution coefficient n_0(ρq_snow/ρ0) for snow [-]"""
+function ν_sno end
+
+""" assumed length scale for water drop or ice crystal [m]"""
+function r0 end
+
+"""coefficient in mass(radius/r0) [kg]"""
+function m0 end
+
+"""coefficient in mass(radius/r0) [-]"""
+function χm end
+
+"""exponent in mass(radius/r0) [-]"""
+function me end
+
+"""exponent in mass(radius/r0) [-]"""
+function Δm end
+
+"""coefficient in area(radius/r0) relation [m2]"""
+function a0 end
+
+"""coefficient in area(radius/r0) [-]"""
+function χa end
+
+"""exponent in area(radius/r0) relation"""
+function ae end
+
+"""exponent in area(radius/r0) [-]"""
+function Δa end
+
+"""coefficient in velocity(radius/r0) [m/s]"""
+function v0 end
+
+"""coefficient in velocity(radius/r0) [-]"""
+function χv end
+
+"""exponent in velocity(radius/r0)"""
+function ve end
+
+"""exponent in velocity(radius/r0) [-]"""
+function Δv end
 
 """drag coefficient for rain drops [-]"""
 function C_drag end
@@ -74,8 +136,8 @@ function C_drag end
 """condensation/evaporation timescale [s]"""
 function τ_cond_evap end
 
-"""sublimation/resublimation timescale [s]"""
-function τ_sub_resub end
+"""sublimation/deposition timescale [s]"""
+function τ_sub_dep end
 
 """autoconversion threshold [-]  ∈(0.5, 1) * 1e-3 """
 function q_liq_threshold end
@@ -83,13 +145,16 @@ function q_liq_threshold end
 """autoconversion timescale [s]  ∈(1e3, 1e4) """
 function τ_acnv end
 
-"""collision efficiency [-]"""
-function E_col end
+"""threshold between ice and snow [m]"""
+function r_ice_snow end
 
-"""ventilation factor coefficient [-]"""
+"""collision efficiency [-]"""
+function E end
+
+"""ventilation factor coefficient for rain or snow [-]"""
 function a_vent end
 
-"""ventilation factor coefficient [-]"""
+"""ventilation factor coefficient for rain or snow [-]"""
 function b_vent end
 
 """thermal conductivity of air [J/m/s/K] """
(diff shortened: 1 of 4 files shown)


src/AtmosParameters.jl
```diff
@@ -13,16 +13,48 @@ AtmosSGS.c_2_KASM(ps::AbstractEarthParameterSet)        = AtmosSGS.c_e2_KASM(ps)
...
 # Microphysics parameters
-Microphysics.MP_n_0(::AbstractEarthParameterSet)          = 8e6 * 2
+Microphysics.n0_rai(::AbstractEarthParameterSet)          = 8e6 * 2
+Microphysics.n0_ice(::AbstractEarthParameterSet)          = 1e7 * 2
+Microphysics.μ_sno(::AbstractEarthParameterSet)           = 4.36 * 1e9
+Microphysics.ν_sno(::AbstractEarthParameterSet)           = 0.63
+
 Microphysics.C_drag(::AbstractEarthParameterSet)          = 0.55
+
+Microphysics.α_rai(::AbstractEarthParameterSet)           = 4/3. * π * CLIMAParameters.Planet.ρ_cloud_liq
```
charleskawczynski, line 25:
```suggestion
Microphysics.α_rai(ps::AbstractEarthParameterSet)           = 4/3. * π * CLIMAParameters.Planet.ρ_cloud_liq(ps)
```
This is a function, so we must call it with the parameter set

src/AtmosParameters.jl
```diff
@@ -13,16 +13,48 @@ AtmosSGS.c_2_KASM(ps::AbstractEarthParameterSet)        = AtmosSGS.c_e2_KASM(ps)
...
+Microphysics.β_rai(::AbstractEarthParameterSet)           = 3.
+Microphysics.γ_rai(::AbstractEarthParameterSet)           = 1. * π
+Microphysics.δ_rai(::AbstractEarthParameterSet)           = 2.
+Microphysics.η_rai(::AbstractEarthParameterSet)           = 0.5
+Microphysics.α_ice(::AbstractEarthParameterSet)           = 4/3. * π * CLIMAParameters.Planet.ρ_cloud_ice
```
charleskawczynski, line 30:
```suggestion
Microphysics.α_ice(ps::AbstractEarthParameterSet)           = 4/3. * π * CLIMAParameters.Planet.ρ_cloud_ice(ps)
```
Same as above

src/Atmos.jl
```diff
@@ -52,21 +52,93 @@ end # module SubgridScale
...
+"""Marshall-Palmer distribution n_0 coeff for rain [1/m4]"""
+function n0_rai end
+
+"""Marshall-Palmer distribution n_0 coeff for ice [1/m4]"""
+function n0_ice end
+
+"""coefficient to compute Marshall-Palmer distribution coefficient n_0 for snow [?]"""
+function μ_sno end
+
+"""coefficient to compute Marshall-Palmer distribution coefficient n_0 for snow [?]"""
```
charleskawczynski, line 101: `[?]` 😱

src/Atmos.jl
```diff
@@ -52,21 +52,93 @@ end # module SubgridScale
...
 
-"""Marshall-Palmer distribution n_0 coeff [1/m4]"""
-function MP_n_0 end
+"""Marshall-Palmer distribution n_0 coeff for rain [1/m4]"""
+function n0_rai end
+
+"""Marshall-Palmer distribution n_0 coeff for ice [1/m4]"""
+function n0_ice end
+
+"""coefficient to compute Marshall-Palmer distribution coefficient n_0 for snow [?]"""
```
charleskawczynski, line 98: `[?]` 😱

src/Atmos.jl
```diff
@@ -83,14 +155,35 @@ function q_liq_threshold end
 """autoconversion timescale [s]  ∈(1e3, 1e4) """
 function τ_acnv end
 
-"""collision efficiency [-]"""
-function E_col end
+"""threshold between ice and snow"""
+function r_ice_snow end
+
+"""collision efficiency for cloud-rain [-]"""
```
charleskawczynski, line 161:
```suggestion
"""collision efficiency for cloud-rain [unitless]"""
```
It might be nice to explicitly say `unitless` in these circumstances. Sorry if I led the way with `[ ]` or `[-]` elsewhere.

src/Atmos.jl
trontrytel: yes... There are 2 parameters that are empirical and have ridiculous  units. I'll discuss this with @tapios in the next microphysics meeting.  The alternative is to have a constant n0 for snow, but thats not what was done in the papers I read so far.

src/Atmos.jl
trontrytel: same as above

src/Atmos.jl
trontrytel: `dimensionless`? Though I prefer the shorter `[-]`...

charleskawczynski: I'm fine with either. `[-]` seems like it could be a placeholder for "to add later" 😬

trontrytel: perfect!

trontrytel:
@vchuravy - could you take a look here. We are trying to add some structure to the free parameters. The goal is to allow nice dispatch when using microphysics functions over different microphysics condensate categories. 

The example use is in my draft PR to ClimateMachine https://github.com/CliMA/ClimateMachine.jl/pull/1067 (the runtest and the Microphysics.jl files)

The main question is: Will it mess things up on the GPU side of things? Happy to explain why I think this is a good idea. Thanks!

src/Atmos.jl
trontrytel: fixed by dividing by rho_0=1kg/m3. Now all params have nice units.

src/AtmosParameters.jl
```diff
@@ -12,18 +12,74 @@ AtmosSGS.c_1_KASM(ps::AbstractEarthParameterSet)        = AtmosSGS.c_a_KASM(ps)*
...
-Microphysics.a_vent(::AbstractEarthParameterSet)          = 1.5
-Microphysics.b_vent(::AbstractEarthParameterSet)          = 0.53
-Microphysics.K_therm(::AbstractEarthParameterSet)         = 2.4e-2
-Microphysics.D_vapor(::AbstractEarthParameterSet)         = 2.26e-5
-Microphysics.ν_air(::AbstractEarthParameterSet)           = 1.6e-5
-Microphysics.N_Sc(ps::AbstractEarthParameterSet)          = Microphysics.ν_air(ps)/Microphysics.D_vapor(ps)
+Microphysics = CLIMAParameters.Atmos.Microphysics
+
+# general
+Microphysics.C_drag(::AbstractEarthParameterSet)  = 0.55
```
jkozdon, line 19: Will this result in type problems further down the stack? (I worry when I see raw decimals)

charleskawczynski: We've decided to wrap everything returned from these functions with `FT`, so that we can make a single global `param_set = EarthParameterSet()` per driver. Otherwise, we'd need to create `param_set = EarthParameterSet{FT}()` in multiple functions/locations. This was a trade-off I discussed with Simon a while back, and we both liked the idea of having a single struct at the top of each driver.

charleskawczynski: Also, the float-type is really orthogonal to the values returned from a parameter set, and it seemed like including `FT` would entangle things that really didn't have to be.

jkozdon: Thanks for the info. Figured you thought of this.

trontrytel:
@charleskawczynski  - I also got a silent node of head from @vchuravy. 

Happy to get it merged, if thats ok with you

tapios:
> @charleskawczynski - I also got a silent nod of head from @vchuravy.
> 
> Happy to get it merged, if thats ok with you

Let me have another quick look over it. Will do it by end of day.

PR #18, opened 2020-05-07
Add parameters

charleskawczynski:
- Adds Mean surface temperature
 - Adds Mean change in virtual temperature

### DIFF
--- a/src/Planet.jl
+++ b/src/Planet.jl
@@ -127,5 +127,9 @@ function orbit_semimaj end
 function TSI end
 """ Mean sea level pressure (Pa) """
 function MSLP end
+""" Mean surface temperature (K) in reference state """
+function T_surf_ref end
+""" Mean vertical virtual temperature drop (K) in reference state """
+function ΔTv_ref end
 
 end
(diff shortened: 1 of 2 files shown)


src/Planet.jl
```diff
@@ -127,5 +127,9 @@ function orbit_semimaj end
 function TSI end
 """ Mean sea level pressure (Pa) """
 function MSLP end
+""" Mean surface temperature (K) """
```
tapios, line 130:
```suggestion
""" Mean surface temperature (K) in reference state"""
```

src/Planet.jl
```diff
@@ -127,5 +127,9 @@ function orbit_semimaj end
...
+function T_surf_ref end
+""" Mean change in virtual temperature (K) """
```
tapios, line 132:
```suggestion
""" Mean vertical virtual temperature drop (K) in reference state"""
```

tapios:
This looks fine. But why not changing the drivers correspondingly in the same go?

charleskawczynski:
> This looks fine. But why not changing the drivers correspondingly in the same go?

This must merge here before it can be used in ClimateMachine (CM) because CLIMAParameters is in CM’s deps, which must be updated to use the latest version. If these parameters were defined in CM, we could do it in one go, but then we’d never be able to split out CM into separate packages

PR #23, opened 2020-05-26
Bump version

charleskawczynski:
Bumping the version so that @trontrytel can use the latest added parameters.


trontrytel:
yeah!

PR #29, opened 2020-06-02
Water property T-dependencies

YujieWangLab:
try to set up global water physical properties, which should be consistent among all Land modules (maybe all CLIMA modules)

### DIFF
--- a/src/Water/density.jl
+++ b/src/Water/density.jl
@@ -1,0 +1,45 @@
+# these constant for used to computed density of air-free water
+const WD_A0 = 999.83952
+const WD_A1 = 16.945176
+const WD_A2 = -7.9870401e-3
+const WD_A3 = -4.6170461e-5
+const WD_A4 =  1.0556302e-7
+const WD_A5 = -2.8054253e-10
+const WD_AC =  0.016897850
+const WD_B0 = 999.85308
+const WD_B1 =  6.32693e-2
+const WD_B2 = -8.523829e-3
+const WD_B3 =  6.943248e-5
+const WD_B4 = -3.821216e-7
+
+
+
+
+"""
+    ρ_water_liq(tem::FT)
+
+Density of air-free pure water `[kg m⁻³]`, given
+- `tem` Water temperature `[K]`
+
+The equation for computing water density (0 to 150 Celcius) is
+```math
+ρ = \\frac{ 999.83952 + 16.945176 \\cdot t - 7.9870401e-3 \\cdot t^2 - 4.6170461e-5 \\cdot t^3 + 1.0556302e-7 \\cdot t^4 - 2.8054253e-10 \\cdot t^5 }{ 1 + 1.6897850e-2 \\cdot t }
+```
+See Kell, George S (1975) Density, thermal expansivity, and compressibility of liquid water from 0 deg. to 150. deg.. Correlations and tables for atmospheric pressure and saturation reviewed and expressed on 1968 temperature scale.
+
+The equation for computing water density (5 to 40 Celcius) is
+```math
+ρ = 999.85308 + 6.32693e-2 \\cdot t - 8.523829e-3 \\cdot t^2 + 6.943248e-5 \\cdot t^3 - 3.821216e-7 \\cdot t^4
+```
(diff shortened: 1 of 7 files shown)


src/Water/density.jl
```diff
@@ -0,0 +1,45 @@
...
+
+"""
+    ρ_water_liq(tem::FT)
+
+Density of air-free pure water `[kg m⁻³]`, given
+- `tem` Water temperature `[K]`
+
+The equation for computing water density (0 to 150 Celcius) is
+```math
+ρ = \\frac{ 999.83952 + 16.945176 \\cdot t - 7.9870401e-3 \\cdot t^2 - 4.6170461e-5 \\cdot t^3 + 1.0556302e-7 \\cdot t^4 - 2.8054253e-10 \\cdot t^5 }{ 1 + 1.6897850e-2 \\cdot t }
```
trontrytel, line 26: wow! Is it really so important to get the water density to such accuracy in the Land model? For clouds we just use a constant. 

And if yes, should we adapt the same in the atmosphere? So that we have consistent (if complicated) water density?

src/Water/latent_heat.jl
```diff
@@ -0,0 +1,45 @@
...
+const SLH_A3 = -0.06       # J kg⁻¹ K⁻³ | Used for specific latent heat
+const SLH_I0 = 2.8341e6    # J kg⁻¹     | Used for specific latent heat
+const SLH_I1 = -290.0      # J kg⁻¹ K⁻¹ | Used for specific latent heat
+const SLH_I2 = -4.0        # J kg⁻¹ K⁻² | Used for specific latent heat
+
+
+
+
+"""
+    LH_v_ice(tem::FT)
```
trontrytel, line 13: Some latent heats are defined in Common/Thermodynamics: 
https://github.com/CliMA/ClimateMachine.jl/blob/766ad0c0354742b08baab5a9d8aef10a511d86c6/src/Common/Thermodynamics/relations.jl#L13-L14

Could they be also used here?

src/Water/vapor_pressure.jl
```diff
@@ -0,0 +1,156 @@
+"""
+    p_sat_water_ice(tem::FT)
+
```
trontrytel, line 3: And here is where the common functions for saturation vapor pressure are defined
https://github.com/CliMA/ClimateMachine.jl/blob/766ad0c0354742b08baab5a9d8aef10a511d86c6/src/Common/Thermodynamics/relations.jl#L568

It would be great to use the same in the clouds and on land.

src/Water/density.jl
cfranken: The ocean model will need this T dependence for sure, should indeed make all of this consistent. I guess it matters mostly for the ocean but why not use the same everywhere.

trontrytel: Fine with me. I think we should use one everywhere if we can

YujieWangLab: It is different for sea water and pure water

src/Water/vapor_pressure.jl
cfranken: It would indeed be good but make Land dependent on the Atmospheric model. Or should we pull Thermodynamics and (and SurfaceFluxes for that matter) out into their own entity (like the Paramaters?)

trontrytel: @charleskawczynski was in the process of pulling Thermodynamics out?

src/Water/density.jl
trontrytel: makes sense...  But then lets make sure that what we use for salty water converges to the not-salty version if salt-> zero

tapios: Keep in mind that the land model deals only with *volume* fractions, hence assume water density to be constant. If you'd want to consider non-constant density, all land equations would have to be formulated as mass conservation equations rather than volume conservation equations. So: water density = constant is what we want in the land model.

src/Water/vapor_pressure.jl
charleskawczynski: Yes, but I think it may be easier to iron out some of Thermo’s test profiles first. We should do this soon, though. Also, it may be easier to increase CLIMAParameter’s precision first, so that we can update the tests in one shot.

PR #30, opened 2020-06-02
decimals->rationals, ints if possible

trontrytel:
@jkozdon - Is that what you had in mind about decimals and ints in parameters?

@charleskawczynski - I noticed that for Prandtl number there was a fractional: https://github.com/CliMA/CLIMAParameters.jl/blob/e28f74631f1aea609b9651c0cbf0eca6e61d8398/src/AtmosParameters.jl#L7 

Looks like those are slow? https://discourse.julialang.org/t/best-practice-for-constant-expressions-1-2-vs-1-0-2-0/17522

### DIFF
--- a/src/AtmosParameters.jl
+++ b/src/AtmosParameters.jl
@@ -1,14 +1,14 @@
 const AtmosSGS = CLIMAParameters.Atmos.SubgridScale
 
 # SubgridScale parameters
-AtmosSGS.C_smag(::AbstractEarthParameterSet)            = 0.21
-AtmosSGS.C_drag(::AbstractEarthParameterSet)            = 0.0011
+AtmosSGS.C_smag(::AbstractEarthParameterSet)            = 21/100
+AtmosSGS.C_drag(::AbstractEarthParameterSet)            = 11/10000
 AtmosSGS.inv_Pr_turb(::AbstractEarthParameterSet)       = 3
-AtmosSGS.Prandtl_air(::AbstractEarthParameterSet)       = 71//100
-AtmosSGS.c_a_KASM(::AbstractEarthParameterSet)          = 0.10
-AtmosSGS.c_e1_KASM(::AbstractEarthParameterSet)         = 0.19
-AtmosSGS.c_e2_KASM(::AbstractEarthParameterSet)         = 0.51
-AtmosSGS.c_1_KASM(ps::AbstractEarthParameterSet)        = AtmosSGS.c_a_KASM(ps)*0.76^2
+AtmosSGS.Prandtl_air(::AbstractEarthParameterSet)       = 71/100
+AtmosSGS.c_a_KASM(::AbstractEarthParameterSet)          = 10/100
+AtmosSGS.c_e1_KASM(::AbstractEarthParameterSet)         = 19/100
+AtmosSGS.c_e2_KASM(::AbstractEarthParameterSet)         = 51/100
+AtmosSGS.c_1_KASM(ps::AbstractEarthParameterSet)        = AtmosSGS.c_a_KASM(ps)*(76/100)^2
 AtmosSGS.c_2_KASM(ps::AbstractEarthParameterSet)        = AtmosSGS.c_e2_KASM(ps)+2*AtmosSGS.c_1_KASM(ps)
 AtmosSGS.c_3_KASM(ps::AbstractEarthParameterSet)        = AtmosSGS.c_a_KASM(ps)^(3/2)
 
@@ -16,7 +16,7 @@ AtmosSGS.c_3_KASM(ps::AbstractEarthParameterSet)        = AtmosSGS.c_a_KASM(ps)^
 const Microphysics = CLIMAParameters.Atmos.Microphysics
 
 # general
-Microphysics.C_drag(::AbstractEarthParameterSet)  = 0.55
+Microphysics.C_drag(::AbstractEarthParameterSet)  = 55/100
 Microphysics.K_therm(::AbstractEarthParameterSet) = 2.4e-2
 Microphysics.D_vapor(::AbstractEarthParameterSet) = 2.26e-5
 Microphysics.ν_air(::AbstractEarthParameterSet)   = 1.6e-5
@@ -31,10 +31,10 @@ Microphysics.τ_sub_dep(::AbstractIceParameterSet)  = 10
 Microphysics.r_ice_snow(::AbstractIceParameterSet) = 62.5 * 1e-6
 Microphysics.n0(::AbstractIceParameterSet)         = 1e7 * 2
 Microphysics.r0(::AbstractIceParameterSet)         = 1e-5
-Microphysics.me(::AbstractIceParameterSet)         = 3.
+Microphysics.me(::AbstractIceParameterSet)         = 3
 Microphysics.m0(
     ps::AbstractEarthParameterSet, ps_ice::AbstractIceParameterSet
-) = 4/3. * π * CLIMAParameters.Planet.ρ_cloud_ice(ps) *
+) = 4/3 * π * CLIMAParameters.Planet.ρ_cloud_ice(ps) *
     Microphysics.r0(ps_ice)^Microphysics.me(ps_ice)
 Microphysics.χm(::AbstractIceParameterSet)         = 1
 Microphysics.Δm(::AbstractIceParameterSet)         = 0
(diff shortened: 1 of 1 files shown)


src/AtmosParameters.jl
```diff
@@ -1,22 +1,22 @@
 const AtmosSGS = CLIMAParameters.Atmos.SubgridScale
 
 # SubgridScale parameters
-AtmosSGS.C_smag(::AbstractEarthParameterSet)            = 0.21
-AtmosSGS.C_drag(::AbstractEarthParameterSet)            = 0.0011
+AtmosSGS.C_smag(::AbstractEarthParameterSet)            = 21/100
```
jkozdon, line 4: Not sure this will make a difference.

I was thinking
```suggestion
AtmosSGS.C_smag(::AbstractEarthParameterSet)            = 21 // 100
```

trontrytel: https://discourse.julialang.org/t/best-practice-for-constant-expressions-1-2-vs-1-0-2-0/17522 - isn't that the slowest option?

jkozdon: Yeah. That surprised me a bit. Though KA inlines everything for us so that should be OK, but what you are doing is fine too.

Advantage of going with rationals is you do not incur the numerical errors until you need to.

jkozdon: This is what surprises me:
```julia
julia> struct Foo end

julia> g(::Foo) = 1//2 * sin(2//3)
g (generic function with 1 method)

julia> f(::T) where T = T(g(Foo()))
f (generic function with 1 method)

julia> @code_llvm f(1.0)

;  @ REPL[3]:1 within `f'
define double @julia_f_16987(double) {
top:
  %1 = alloca [2 x i64], align 8
; ┌ @ REPL[2]:1 within `g'
; │┌ @ rational.jl:44 within `//'
; ││┌ @ rational.jl:21 within `Rational' @ rational.jl:15
     call void @julia_divgcd_16988([2 x i64]* noalias nocapture nonnull sret %1, i64 2, i64 3)
; │││┌ @ tuple.jl:63 within `indexed_iterate' @ tuple.jl:63
      %2 = getelementptr inbounds [2 x i64], [2 x i64]* %1, i64 0, i64 0
; ││││ @ tuple.jl:63 within `indexed_iterate'
      %3 = getelementptr inbounds [2 x i64], [2 x i64]* %1, i64 0, i64 1
; │└└└
; │┌ @ special/trig.jl:53 within `sin'
; ││┌ @ float.jl:277 within `float'
; │││┌ @ rational.jl:91 within `AbstractFloat'
; ││││┌ @ float.jl:277 within `float'
; │││││┌ @ float.jl:262 within `AbstractFloat'
; ││││││┌ @ float.jl:60 within `Float64'
         %4 = load i64, i64* %2, align 8
         %5 = sitofp i64 %4 to double
         %6 = load i64, i64* %3, align 8
         %7 = sitofp i64 %6 to double
; ││││└└└
; ││││┌ @ float.jl:407 within `/'
       %8 = fdiv double %5, %7
; ││└└└
    %9 = call double @julia_sin_16990(double %8)
; │└
; │┌ @ promotion.jl:312 within `*' @ float.jl:405
    %10 = fmul double %9, 5.000000e-01
; └└
  ret double %10
}
```

Why isn't constant propagation working right? This is closer to what we do in practice.

jkozdon: But ` 0.21` vs `21 / 100` will make no difference:
```julia
julia> 0.21 === 21 / 100
true
```

jkozdon: If we mark all the functions as `pure` we would be OK too:
```julia
julia> struct Foo end

julia> Base.@pure g(::Foo) = 1//2 * sin(2//3)
g (generic function with 1 method)

julia> f(::T) where T = T(g(Foo()))
f (generic function with 1 method)

julia> @code_llvm f(1.0)

;  @ REPL[27]:1 within `f'
define double @julia_f_17034(double) {
top:
  ret double 0x3FD3C9AF78209765
}
```

trontrytel: I was confused about why it would

jkozdon: Though I do not understand why this doesn't fix the broadcast in the discourse example....

```julia
julia> Base.@pure f1(x) = 1/2 * sin(2/3 * x)
f1 (generic function with 1 method)

julia> Base.@pure f2(x) = 1.0/2.0 * sin(2.0/3.0 * x)
f2 (generic function with 1 method)

julia> Base.@pure f3(x) = 1//2 * sin(2//3 * x)
f3 (generic function with 1 method)

julia> using BenchmarkTools

julia> @btime f1(0.5);
  0.018 ns (0 allocations: 0 bytes)

julia> @btime f2(0.5);

  0.018 ns (0 allocations: 0 bytes)

julia> @btime f3(0.5);

  0.018 ns (0 allocations: 0 bytes)

julia> x = rand(10000);

julia> @btime f1.($x);

  49.301 μs (2 allocations: 78.20 KiB)

julia> @btime f2.($x);

  49.248 μs (2 allocations: 78.20 KiB)

julia> @btime f3.($x);
  191.688 μs (2 allocations: 78.20 KiB)
```

jkozdon: > Though I do not understand why this doesn't fix the broadcast in the discourse example....

Though we won't broadcast there, so maybe no problem for us. Our situation is more like:
```
julia> struct Foo end

julia> Base.@pure f1(::Foo) = 0.5
f1 (generic function with 2 methods)

julia> Base.@pure f2(::Foo) = 1 // 2
f2 (generic function with 2 methods)

julia> g1(a, x) = x .+ eltype(x)(f1(a))
g1 (generic function with 1 method)

julia> g2(a, x) = x .+ eltype(x)(f2(a))
g2 (generic function with 1 method)

julia> using BenchmarkTools

julia> a = Foo()
Foo()

julia> @btime g1($a, 0.5);

  0.018 ns (0 allocations: 0 bytes)

julia> @btime g2($a, 0.5);

  0.018 ns (0 allocations: 0 bytes)

julia> x = rand(100);


julia> @btime g1($a, $x);

  76.714 ns (1 allocation: 896 bytes)

julia> @btime g2($a, $x);
  77.067 ns (1 allocation: 896 bytes)
```

jkozdon: And... if you did want to broadcast that's fine with this style function as well with the `@pure`:
```julia
julia> struct Foo end

julia> Base.@pure f1(::Foo) = 0.5
f1 (generic function with 1 method)

julia> Base.@pure f2(::Foo) = 1 // 2
f2 (generic function with 1 method)

julia> g1(a, x::T) where T = x + T(f1(a))
g1 (generic function with 1 method)

julia> g2(a, x::T) where T = x + T(f2(a))
g2 (generic function with 1 method)

julia> using BenchmarkTools

julia> a = Foo();

julia> @btime g1($a, 0.5);
  0.018 ns (0 allocations: 0 bytes)

julia> @btime g2($a, 0.5);
  0.018 ns (0 allocations: 0 bytes)

julia> x = rand(100);

julia> a = fill(Foo(), length(x));

julia> @btime g1.($a, $x);
  77.298 ns (1 allocation: 896 bytes)

julia> @btime g2.($a, $x);
  77.878 ns (1 allocation: 896 bytes)
```

trontrytel: So whats our strategy? Should I mark the functions as `pure` and switch to `//` where possible?

simonbyrne: I don't think we should bother with rationals: unless we want to use `BigFloat`s, any increase in accuracy by avoiding the double rounding of Rational => Float64 => Float32 is pretty minor. `@pure` is somewhat dangerous, and when we add functionality for modifiable parameter sets we will be violating its assumptions.

jkozdon: > don't think we should bother with rationals

Fair enough.

jkozdon:
> @jkozdon - Is that what you had in mind about decimals and ints in parameters?

Concerning `Int`s though, I think that for parameters `1` is better than `1.`.

trontrytel:
> > @jkozdon - Is that what you had in mind about decimals and ints in parameters?
> 
> Concerning `Int`s though, I think that for parameters `1` is better than `1.`

I'll change that

src/AtmosParameters.jl
jkozdon: > unless we want to use `BigFloat`s

Would you ever want to run in `DoubleFloat`?

charleskawczynski: My 2 cents: I'd avoid rationals for now

trontrytel:
> These changes are fine by me. Let's update the title since we're not converting to rationals.

Sure. But wait. I'll remove the unneccessary changes to decimals. And just keep the ints

PR #31, opened 2020-06-02
add surface tension and viscosity parameters

Yujie-W:


tapios:
These look like empirical parameters in a parameterization, which may go into CliMAParameters. The planet parameters are meant to be true constants for a planet. 

If you move this to the adjustable parameters, please choose variable names that are descriptive and clear (some verbosity is ok--it would be good if it's easy for people to understand what the parameters are about).

charleskawczynski:
I'm not sure that `Water` exists at the correct hierarchy level. This may need to be fixed/updated in `src/` and `docs/`. Can/may these properties be used outside of the land model? Also, if it's only part of the land model then should we enter this into `CLIMAParameters.Land.Water`?

trontrytel:
> I'm not sure that `Water` exists at the correct hierarchy level. This may need to be fixed/updated in `src/` and `docs/`. Can/may these properties be used outside of the land model? Also, if it's only part of the land model then should we enter this into `CLIMAParameters.Land.Water`?

I think this could work. We could have `WaterParameters.jl` with all water properties. This would be then used in Atmos, Land and Ocean

PR #45, opened 2021-03-01
add orbital parameters needed in Insolation.jl

claresinger:


src/Planet/Planet.jl
```diff
@@ -127,6 +132,16 @@ function year_anom end
 function orbit_semimaj end
 """ Total solar irradiance (W/m``^2``) """
 function TSI end
+""" J2000 epoch """
```
tapios, line 135:
```suggestion
""" Time of  epoch (J2000)"""
```

src/Planet/Planet.jl
```diff
@@ -127,6 +132,16 @@ function year_anom end
...
+function epoch end
+""" mean anomaly at the epoch (radians) """
+function M_epoch end
```
tapios, line 138: Consider more verbose variable names for all, e.g., `mean_anom_0`, `obliquity`, `lon_perihel` etc.

src/Planet/Planet.jl
```diff
@@ -127,6 +132,16 @@ function year_anom end
...
+""" orbital obliquity at the epoch (radians) """
+function γ_epoch end
+""" longitude of perihelion at the epoch (radians) """
```
tapios, line 141:
```suggestion
""" longitude of perihelion at the epoch (radians), measured relative to vernal equinox (i.e., the longitude of perihelion is the angle subtended at the Sun by the orbital arc from vernal equinox to perihelion)."""
```
(Good to be explicit here, because there are various definitions of the longitude of perihelion in use.)

charleskawczynski (commented):
It might be good to increase the precision of these parameters, if possible. I think after that, and squashing, we'll be ready to merge

charleskawczynski (approved):

charleskawczynski:
Looks good!

PR #48, opened 2021-03-05
Add EDMF parameters.

ilopezgp:
This PR adds the EDMF parameters, which will then be removed from ClimateMachine.jl.


ilopezgp:
@charleskawczynski I think this is ready

charleskawczynski (approved):

PR #50, opened 2021-07-14
Add water surface tension coefficient for AerosolActivation.jl

trontrytel:


charleskawczynski (approved):
LGTM. Do you want to bump the version in a subsequent commit so that you can start using them?

trontrytel:
That would be great!

We will also have a bigger bunch of parameters incoming that describe the aerosol properties soon. But we can bump the version again then.

Issue #43, opened 2020-11-04
Ocean density must be referred to as a reference density


glwagner:
For detail-oriented folks, the variable

https://github.com/CliMA/CLIMAParameters.jl/blob/f779db6fa86f0bfce9de4ee31757fed1fe8b75ed/src/Planet/planet_parameters.jl#L36

is somewhat stressful, as the ocean density is not constant.

This quantity is really a reference density that must be defined as part of the Boussinesq approximation; it's important to identify it as a "reference" (and therefore a parameter) rather than an absolute value.

It's appropriate to include this parameter here, because it ends up appearing in a few unrelated places in ocean codes (and also user's scripts). But I think it should be called `ocean_reference_density` so as to not cause unwarranted stress for users that want to carefully manage their science.

charleskawczynski:
Sounds like a good idea. In the atmosphere, we append `ref` (e.g., `T_min_ref`) to indicate reference values.

glwagner:
For sure. That can work for mathematical notation form. Going along with #41 I think we should use the literal word `reference` to indicate reference values for English variable names.

PR #51, opened 2021-07-27
Add some planet parameters

charleskawczynski:
These parameters are from "Large-eddy simulation in an anelastic framework with closed water and entropy balances" (Pressel et. al.), Table 1. This is needed for [TurbulenceConvection.jl's #43](https://github.com/CliMA/TurbulenceConvection.jl/pull/43).

### DIFF
--- a/src/Planet/Planet.jl
+++ b/src/Planet/Planet.jl
@@ -34,6 +34,9 @@ export molmass_dryair,
     e_int_i0,
     press_triple,
     surface_tension_coeff,
+    entropy_dry_air,
+    entropy_water_vapor,
+    entropy_reference_temperature,
     ρ_ocean,
     cp_ocean,
     planet_radius,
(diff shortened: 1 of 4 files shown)


src/Planet/Planet.jl
```diff
@@ -114,6 +117,23 @@ function e_int_i0 end
...
+on the reference data given in Lemmon et al. [2000], and
+the standard entropy value for water vapor is based on the
+reference data given in Chase [1998].
+=#
+
+""" Entropy of dry air J / (kg K)"""
+function entropy_dry_air end
+""" Entropy of water vapor J / (kg K)"""
+function entropy_water_vapor end
+""" Standard temperature (K)"""
```
tapios, line 134: Maybe better to make explicit that this is the reference temperature for entropies (we have several other reference temperatures). E.g., 
```suggestion
""" Entropy reference temperature (K)"""
```

src/Planet/Planet.jl
```diff
@@ -114,6 +117,23 @@ function e_int_i0 end
...
+function standard_temperature end
```
tapios, line 135:
```suggestion
function entropy_reference_temperature end
```

tapios (approved):

charleskawczynski:
Thanks for the review!

PR #54, opened 2021-08-25
Remove microphysics types

charleskawczynski:
We ran into some issues as TurbulenceConvection.jl began using CLIMAParameters more heavily-- specifically with the microphysics parameters:

For planet parameters, we might use a runtime behavior like:
```julia
import CLIMAParameters
CLIMAParameters.Planet.grav(ps::AbstractEarthParameterSet) = ps.grav
```
This same approach for certain microphysics parameters is awkward:
```julia
CLIMAParameters.Planet.Microphysics.E(::AbstractLiquidParameterSet, ::AbstractRainParameterSet) = ?
```
The issue is, which struct should hold the value if we want it tuned? This seems arbitrary and generally problematic with this design.

This PR makes _all_ (microphysics) methods accept a single parameter set, and we can locally define extra helper methods to dispatch in convenient ways:

```julia
# How to add convenience methods for handling combinations

abstract type Phase end
struct Liquid <: Phase end
struct Rain   <: Phase end
struct Snow   <: Phase end

E(earth::AbstractEarthParameterSet, ::Liquid, ::Rain) = CLIMAParameters.Planet.Microphysics.E_liq_rai(earth)
E(earth::AbstractEarthParameterSet, ::Liquid, ::Snow) = CLIMAParameters.Planet.Microphysics.E_liq_sno(earth)

struct EarthParamSet <: AbstractEarthParameterSet end

earth = EarthParamSet()

# locally, in the implementation, we'll have:
E_liq_rai = E(earth, Liquid(), Rain())
E_liq_sno = E(earth, Liquid(), Snow())
```


trontrytel (approved):
As far as I can check with my human eyes, it looks ok. I guess we will see if there are any typos once we update the dependency in CloudMicrophysics.jl

charleskawczynski:
Thank you for taking a look @trontrytel!

charleskawczynski:
Going to move forward with this.

PR #56, opened 2021-12-10
bump version, use julia 1.7

trontrytel:
This PR:
- bumps the repo version (one of the parameter names was changed in the last PR, so it was an interface change)
- tries using Julia 1.7 in CI


charleskawczynski (approved):
LGTM, let’s just squash first

PR #61, opened 2022-01-12
Update EDMF parameter notation.

ilopezgp:
Updates the naming convention for one of the EDMF parameters, following the parameter table in Cohen et al (2020). It also fixes the default value of static_stab_coefficient `c_b`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CLIMAParameters"
 uuid = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 authors = ["Charles Kawczynski <kawczynski.charles@gmail.com>"]
-version = "0.3.0"
+version = "0.4.0"
 
 [deps]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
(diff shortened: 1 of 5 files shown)


yairchn (approved):
LGTM - I think there should a corresponding change in TC that merge simultaneously? 
or we can overwrite the parameter in TC.jl until correspond changes take place

charleskawczynski (commented):
LGTM, this is a breaking change, so we'll need to bump the minor version for the next release.

ilopezgp:
For now, I included `c_γ` in the `src/ClimaParams.jl` script in TC.jl, so I was able to merge the change. Is this still a breaking change after I removed `c_t` from the TC.jl codebase @charleskawczynski ?

charleskawczynski:
> For now, I included `c_γ` in the `src/ClimaParams.jl` script in TC.jl, so I was able to merge the change. I think this will not break TC.jl after I removed `c_t` from the TC.jl codebase @charleskawczynski ?

Yes, but will it break any other packages? Several other packages use CLIMAParameters (e.g., CloudMicrophysics.jl).

> My idea is to first create a release of TC.jl, then of ClimaParameters.jl, so that the update does not break the code of users after they update TC.jl. But yes this will require a breaking release.

For pre 1.0.0, nothing will automatically break if we bump the minor version (0.2.3 = major.minor.patch) because `] up` will only auto update the patch version. For pre 1.0.0, Minor versions need to be updated manually, allowing us to merge / release this PR, then merge/release TC.jl

PR #63, opened 2022-02-11
[WIP] pilot parameter refactor (b) 

odunbar:
## Purpose
Generalize away from alias => `::Dict`. Also we can now write **re-readable** log files

## In this PR
- Documentation! See  the latest documentor build below.
- New wrapper ```ParamDict{FT}```around dictionary read from toml. Currently parametric on one type of all the values, provided when it is built (default is `Float64`). Construction:
```julia
create_parameter_struct(override_filepath, default_filepath; dict_type="alias",value_type=Float64) # builds with override and custom default provided
create_parameter_struct(override_filepath ; dict_type="alias",value_type=Float64) # builds, overriding the default from CLIMAParameters
create_parameter_struct(; dict_type="alias",value_type=Float64) # builds using only default in CLIMAParameters
```
- Automated parameter logging on reading parameter values: (can do multiple parameters at once). At the call of this function, the type is enforced on the return values.
```julia 
get_parameter_values!(pd::ParamDict, names, component; log_component=true) #logs component in pd
get_parameter_values(pd::ParamDict,names) # no logging in pd
```
- writing the re-readable log file in `log_parameter_information(pd::ParamDict,filepath)`. Example parameter tested in `CloudMicrophysics.jl`:
```toml
# initial parameter file
[molar_mass_water]
alias = "molmass_water"
value = 0.01801528
type = "float"
```
In this test, a copy of the parameters was created and stored in (1) Thermodynamics, (2) the general CloudMicrophysics, and (3) the specific 1-moment-based Microphysics scheme. Indeed the logfile shows this information below
```toml
# log file
[molar_mass_water]
alias = "molmass_water"
value = 0.01801528
type = "float"
used_in = ["Thermodynamics", "CloudMicrophysicsParameters", "Microphysics_1M_Parameters"]
```
The `log_parameter_information(pd::ParamDict,filepath; warn_else_error="warn")` will also throw warnings if a parameter is in the override file, but unused in the simulation. This works as parameters from the override file are checked for whether they have gained a "used_in" tag, if not found then an this is logged as a warning or error
- From PR #65, reading of array-type parameters
- From PR #68, reading and writing for calibration tools
### TODO
- still using alias based structs
- still based on float, and array-of-float parameters

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Charles Kawczynski <kawczynski.charles@gmail.com>"]
 version = "0.4.0"
 
 [deps]
+DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,10 @@ authors = ["Charles Kawczynski <kawczynski.charles@gmail.com>"]
 version = "0.4.0"
 
 [deps]
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+EnsembleKalmanProcesses = "aa8a2aa5-91d8-4396-bcef-d4f2ec43552d"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/src/file_parsing.jl
+++ b/src/file_parsing.jl
@@ -125,7 +125,7 @@ end
 
 
 
-log_component!(param_set::ParamDict{FT},names,component) = log_component!(param_set.data,names,component,param_set.dict_type)
+log_component!(param_set::ParamDict{FT},names,component) where {FT} = log_component!(param_set.data,names,component,param_set.dict_type)
 
 function log_component!(data::Dict,names,component,dict_type)
     if dict_type == "alias"
@@ -156,7 +156,7 @@ function log_component!(data::Dict,names,component,dict_type)
     end
 end
 
-get_values(param_set::ParamDict{FT}, names) = get_values(param_set.data, names, param_set.dict_type)
+get_values(param_set::ParamDict{FT}, names) where {FT} = get_values(param_set.data, names, param_set.dict_type)
 
 function get_values(data::Dict, names, dict_type)
 
@@ -177,7 +177,7 @@ function get_values(data::Dict, names, dict_type)
     return ret_values
 end
 
-function get_parameter_values!(param_set::ParamDict{FT}, names, component; log_component=true)
+function get_parameter_values!(param_set::ParamDict{FT}, names, component; log_component=true) where {FT}
     names_vec = (typeof(names) <: AbstractVector) ? names : [names]
     
     if log_component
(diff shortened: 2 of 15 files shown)


src/file_parsing.jl
```diff
@@ -4,80 +4,175 @@ function parse_toml_file(filepath)
...
+            for (kkey,vval) in val # as val is a Dict too
+                merged_struct.data[key][kkey] = vval
+            end
+        end
+    end
+    return merged_struct
+end
+
+
+function create_parameter_struct(path_to_override, path_to_default; dict_type="alias")
```
charleskawczynski, line 141: We should probably avoid `try-catch` here.

odunbar:
Sorry @simonbyrne it has taken me a while - but i've only just seen this request. I shall discuss this on Monday with @bielim , who will be looking at the calibration side of this, and thus the mechanism of overriding in practice.

The only thing i can see currently is that this will force e.g. priors, to be defined in the CLIMAParameters defaults file and thus they cannot be changed for easy prototyping. 

An example of an alternative (not necessarily what i prefer) is softer, by warning/sending an error if a parameter in the overrride file was not used.

odunbar:
@simonbyrne I've decided to stick with warnings/errors for the parameters on logging for now. 

This seems to accomplish the task of notifying users of any issues with typos (or likewise if source code changes), I've tried to make the error/warning messages clear. i.e the likely cause is a typo. (The only way around this message names a parameter as another parameter - unlikely with descriptive names).

The organic question of which parameters will end up being defaulted is still in flux, so I would like to leave as little constrained to this list as possible as the climate model develops

Project.toml
```diff
@@ -4,6 +4,10 @@ authors = ["Charles Kawczynski <kawczynski.charles@gmail.com>"]
 version = "0.4.0"
 
 [deps]
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
+EnsembleKalmanProcesses = "aa8a2aa5-91d8-4396-bcef-d4f2ec43552d"
```
charleskawczynski, line 8: Is it necessary that we add all of these dependencies?

odunbar: So we have file readers from the clima side, (which don't need these) and file readers from the calibration side (which do). This is because the code is effectively this: 
```f("text_defining_EKP_prior") = instantiation_of_an_EKP_prior.``` 
Currently all the code for the priors is in EKP - so for now we would have to. 

The alternative, would be to move the code to read and write toml files for calibration purposes out to EKP, but we have written this code here for now. What do you think?

File: src/file_parsing.jl [hunk 3]
```diff
@@ -1,83 +1,327 @@
...
+    iterate_alias,
+    log_component!,
+    get_values,
+    get_parameter_values!,
+    get_parameter_values,
+    check_override_parameter_usage,
+    write_log_file,
+    log_parameter_information,
+    merge_override_default_values,
+    create_parameter_struct
```
charleskawczynski, line 16:
```suggestion
```

let's not export to encourage qualifying types and methods.

src/file_parsing.jl
```diff
@@ -1,83 +1,327 @@
...
+    dict_type::String
+    "either a nothing, or a dictionary representing an override parameter TOML file"
+    override_dict::Union{Nothing,Dict}
+end
+
+"""
+    parse_toml_file(filepath)
+
+use a TOML parser to read TOML file at `filepath`.
+"""
```
charleskawczynski, line 45: Can we just inline (and delete) this method?

src/file_parsing.jl
```diff
@@ -1,83 +1,327 @@
...
 function parse_toml_file(filepath)
     return TOML.parsefile(filepath)
 end
 
+"""
+    get_parametric_type(::ParamDict{FT}) where {FT}
 
+obtains the type `FT` from `ParamDict{FT}`.
+"""
+get_parametric_type(::ParamDict{FT}) where {FT} = FT
```
charleskawczynski, line 55:
```suggestion
float_type(::ParamDict{FT}) where {FT} = FT
```
And fix docs accordingly

src/file_parsing.jl
```diff
@@ -1,83 +1,327 @@
...
+        end
+    end
+end
+
+"""
+    write_log_file(param_set::ParamDict{FT}, filepath) where {FT}
+
+Writes a log file of all used parameters of `param_set` at the `filepath`. This file can be used to rerun the experiment.
+"""
+function write_log_file(param_set::ParamDict{FT}, filepath) where {FT}
```
charleskawczynski, line 226:
```suggestion
function write_log_file(param_set::ParamDict{FT}, filename) where {FT}
```

Project.toml
charleskawczynski: I think I'd prefer option 1, @simonbyrne, thoughts?

src/file_parsing.jl
charleskawczynski: Oh wait, is this not always the float type?

odunbar: It will be floats. Currently we have not had need for other types (except for arrays of floats) 

(PS I'll address your other comments in due course!)

PR #65, opened 2022-03-03
Enable get_parameter_values to deal with array-valued parameters

bielim:
- `get_parameter_values` extracts the requested parameter values and returns
  them as an array. The returned array contains the values corresponding
  to the given list of parameter names (unless a single scalar  parameter is
  requested, in which case the function returns a float)
- The return type of `get_parameter_values` is determined by the type of the 
  `ParamStruct` the parameter values are extracted from. If e.g. the `ParamStruct`
  is of type `Float64`, then `get_parameter_values` returns parameter values 
  of type `Float64`. Note that the list of parameter names passed to 
  `get_parameter_values` can include both array-valued and scalar-valued
  parameters.
- E.g., given a list of parameter names ["scalar_param", "array_param"] with
  `scalar_param` having a scalar value and `array_param` being array-valued,
  the first element of the returned array will be a float, and the second
  element will be an array of floats
- Added more tests to `toml_consistency.jl` to check if parsing, extracting,
  and writing of array-valued parameters works. Also added a test for
  `merge_override_default_values`


src/array_parameters.toml
```diff
@@ -0,0 +1,14 @@
+[array_parameter_1]
```
odunbar, line 1: This needs to be in `test/` not `src/`

src/file_parsing.jl
```diff
@@ -80,22 +80,22 @@ function log_component!(data::Dict,names,component,dict_type)
...
 end
 
-get_values(param_set::ParamDict{FT}, names) where {FT} = get_values(param_set.data, names, param_set.dict_type, get_parametric_type(param_set))
+get_values(param_set::ParamDict{FT}, names) where {FT} = get_values(param_set.data, names, param_set.dict_type)
 
-function get_values(data::Dict, names, dict_type, ret_values_type)
+function get_values(data::Dict, names, dict_type)
     
-    ret_values = ret_values_type[]
+    ret_values = []
```
odunbar, line 87: We need to enforce the `Float` type somewhere: before, we created e.g. a `Float64[]` or `Float32[]` array. Then upon reading it converts the read value to `Float64` or `Float32`. 

If you remove this here, you need to enforce it elsewhere.

odunbar: e.g. within the `push!(...)` operation

test/toml_consistency.jl
```diff
@@ -89,7 +91,71 @@ universal_constant_aliases = [
...
+    end
     
+    # Write parameters to log file
+    logfilepath = joinpath(@__DIR__, "log_file_test_2.toml")
+    CP.write_log_file(param_set, logfilepath)
 
+    # `param_set` and `full_param_set` contain different values for the
+    # `gravitational_acceleration` parameter. The merged parameter set should
+    # contain the value from `param_set`.
+    full_param_set = CP.create_parameter_struct(dict_type="name")
```
odunbar, line 153: The `merge_override_default_values` is already called within 
`param_set = create_parameter_struct(path_to_array_params,...)`

you shouldn't need to call the default outside - (but it is fine to test it)

odunbar (approved):
This looks great thanks! 

Before merging I will test it across with `Thermodynamics` package to see if anything comes up

odunbar:
OK - I can confirm this passes tests in Thermodynamics:
https://buildkite.com/clima/thermodynamics-ci/builds/638

odunbar:
We can just merge into my branch

PR #68, opened 2022-03-10
[WIP] Read and write UQ parameters

bielim:
- [x] Read UQ parameters from TOML file. UQ parameters come with prior distributions and constraints (see `ParameterDistributionStorage`), which have to be reconstructed from the parsed TOML file
- [x] Write UQ parameters to file, 
- [x] Implement "broadcasting solution" for defining priors of array-valued parameters (use case: a 100-dim parameter of weights of a neural network, each of which has the same prior distribution. Users should not have to write down 100 parameter distributions in the parameter toml file for cases like this one - rather, we want them to be able to write down a single distribution that gets broadcast and applied to each of the 100 weights).
- [x] Clean up: consistent formatting, consistent naming of variables

-----
Current implementation for constructing parameter distributions for parameters read from file `uq_parameters.toml`: 
```
import CLIMAParameters
const CP = CLIMAParameters
param_set = CP.parse_toml_file("uq_test_parameters.toml")

# get `ParameterDistribution` for the parameter named "uq_param_1" (a 10d parameter of type `Parameterized`)
pd1 = CP.get_parameter_distribution(param_set, "uq_param_1")
# get `ParameterDistribution` for the parameter named "uq_param_5" (a 4-dim parameter of type `Samples`)
pd5 = CP.get_parameter_distribution(param_set, "uq_param_5")
# get `ParameterDistribution`s for parameters "uq_param_1" and "uq_param_5
pd = CP.get_parameter_distribution(param_set, ["uq_param_1", "uq_param_5"])

### CODE REVIEW

File: src/file_parsing_uq.jl [hunk 1]
```diff
@@ -0,0 +1,190 @@
...
+            return dist_type(dist_args)
+        else
+            throw(error("Unknown distribution type ", dist_type))
+        end
+    end
+end
+
+
+function construct_2d_array(expr)
+    """
```
odunbar, line 129: Thanks for doing all the docstrings! However could you move them above the functions?

bielim: Done!

odunbar:
@bielim 
To run the tests i needed to 
1. add `include("uq_parameters.jl")` into `test/runtests.jl`
2. add `Distributions` and `EnsembleKalmanProcesses` into `CLIMAParameters` project toml

test/uq_parameters.jl
```diff
@@ -0,0 +1,224 @@
...
+    param_names = ["uq_param_2", "uq_param_4", "uq_param_5"]
+    pd = CP.get_parameter_distribution(param_dict, param_names)
+
+    # Save the parameter dictionary and re-load it. 
+    logfile_path = joinpath(@__DIR__,"log_file_test_uq.toml")
+    CP.write_log_file(param_dict, logfile_path)
+
+    # Read in log file as new parameter file and rerun test.
+    param_dict_from_log = CP.parse_toml_file(logfile_path)
+    rm(logfile_path) # clean up
```
odunbar, line 95: move this below the test in case it fails partway through, or even leave it as it is a log-file. In general these will overwrite existing logs I believe.

test/uq_parameters.jl
```diff
@@ -0,0 +1,224 @@
...
+    uq_param_path = joinpath(@__DIR__,"uq_test_parameters.toml")
+    param_set = CP.create_parameter_struct(
+        uq_param_path,
+        param_path,
+        dict_type="name"
+    )
+
+    # Extract the UQ parameters from the joint set of the parameters from
+    # param_path and those from uq_param_path
+    uq_param_names = CP.get_UQ_parameters(param_set)
```
odunbar, line 122: I'm not sure this is the right use-case
```julia
param_set = CP.create_parameter_struct(
        uq_param_path,
        param_path,
        dict_type="name"
    )
uq_param_names = CP.get_UQ_parameters(param_set)
```
I think we should have 
1. a single file with both uq and other parameters in it
2. a toml parsing function in `file_parsing_uq.jl` to load up the parameters from the toml file 
Currently this code uses the `create_param_struct` from CLIMAParameters which does additional loading and overriding of defaults.

odunbar: If you are only using this in the test case - that is 'ok' but could you perhaps make it more explicit / provide the `file_parsing_uq.jl` reader for the toml file?

test/uq_parameters.jl
```diff
@@ -0,0 +1,224 @@
...
+        initial_ensemble,
+        y_obs,
+        Γy,
+        Inversion(),
+        rng=rng)
+
+    # EKS iterations
+    for i in 1:N_iter
+        params_i = get_u_final(eksobj)
+        G_n = [G(params_i[1, i],
```
odunbar, line 191: see the `batch` function in ParameterDistributions, which should provide the `[1,2,3:6,7:8,...]` etc.

odunbar (commented):
Thanks for all the work! This looks great

 If this satisfies the CalibrateEDMF use cases then I think my only issue is the loading from file should have it's own function in `file_parsing_uq.jl`

test/uq_parameters.jl
bielim: Yes, I'll change this - it's a good opportunity to showcase `batch` (of whose existence I didn't know yet when I first wrote that ensemble update :-))!

test/uq_parameters.jl
bielim: I removed the remove command

test/uq_parameters.jl
bielim: I added a parse function `read_parameters` to `file_parsing_uq.jl`. `read_parameters` only takes a single argument, the path to the parameter file (a .toml file) that is to be read. No override / default combinations and distinctions; the assumption is that all parameters (parameters to be calibrated and other ones) live in a single file.

odunbar: Great! I'll take a look

test/uq_test_parameters.toml
```diff
@@ -0,0 +1,82 @@
+# The six parameters below are to be used for calibration / uncertainty
```
odunbar, line 1: seven

test/uq_test_parameters.toml
```diff
@@ -0,0 +1,82 @@
...
+# e.g., "uq_param_6_(1)", "uq_param_6_(2)", etc.
+[uq_param_6]
+prior = "repeat([Parameterized(Gamma(2.0, 1.0))], 3)"
+constraint = "repeat([bounded_above(9.0)], 3)"
+type = "float"
+
+# A repeated constraint can also be combined with a multivariate distribution.
+[uq_param_7]
+prior = "Parameterized(MvNormal(3, 2.0))"
+constraint = "repeat([no_constraint()], 3)"
```
odunbar, line 43: I presumed that 
`constraint = "repeat([no_constraint()], 3)"` would interpret as
`[[no_constraint()], [no_constraint()], [no_constraint()] ]` 
(which is what appears to happen with `UQ_param_6`)

And `constraint = "repeat(no_constraint(), 3)"` would be 
`[no_constraint(), no_constraint(), no_constraint()] ` 
But this appears to give an error for `UQ_param_7`?

test/uq_test_parameters.toml
```diff
@@ -0,0 +1,82 @@
...
+constraint = "no_constraint()"
+type = "float"
+
+[uq_param_2]
+prior = "Parameterized(Gamma(5.0, 2.0))"
+constraint = "bounded_below(3.0)"
+type = "float"
+
+[uq_param_3]
+prior = "Parameterized(MvNormal(4, 0.1))"
```
odunbar, line 14: I see you do not have any value types here. are these ignored (I'd recommend this behaviour)

test/uq_test_parameters.toml
```diff
@@ -0,0 +1,82 @@
...
+
+[uq_param_4]
+prior = "Samples([1.0 3.0 5.0 7.0; 9.0 11.0 13.0 15.0])"
+constraint = "[bounded(10.0, 15.0), bounded_below(-1.0)]"
+type = "float"
+
+[uq_param_5]
+prior = "Samples([1.0 3.0; 5.0 7.0; 9.0 11.0; 13.0 15.0])"
+constraint = "[no_constraint(), no_constraint(), bounded_below(-2.0), bounded_above(100.0)]"
+type = "float"
```
odunbar, line 26: presumably these are arrays not floats? (I guess this means that you ignore this field - perhaps it can be removed here, or at least corrected)

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,686 @@
...
+            Defaults to "constrained" but can also be set to "unconstrained"
+"""
+function save_parameter_ensemble(
+    param_array::Array{FT, 2},
+    param_distribution::ParameterDistribution,
+    default_param_data::Dict,
+    save_path::AbstractString,
+    save_file::AbstractString,
+    iter::Union{Int, Nothing}=nothing,
+    save_as::AbstractString="constrained") where {FT}
```
odunbar, line 452: Is there a use-case for this optional argument?

odunbar (commented):
Just a couple of minor points. Mainly around the parameter file parsing

odunbar:
Can you also add
`Random` and `LinearAlgebra` into the `Project.toml`
This will allow 
```
julia --project
] 
test
```
to work

src/file_parsing_uq.jl
```diff
@@ -0,0 +1,686 @@
...
+        # parameters
+        if typeof(names) <: AbstractString
+            # The broadcast parameter is the only building block of the
+            # parameter distribution
+            return ParameterDistribution(
+                prior[1],
+                constraint[1],
+                identity.(param_names))
+        else
+            # The ParaeterDistribution consists of a broadcast parameter and at
```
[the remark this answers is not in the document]
bielim: Fixed

test/uq_test_parameters.toml
bielim: Yes, this was somewhat inconsistent. I changed the specification of repeated constraints in the toml files to

```
[uq_param_6]
prior = "repeat([Parameterized(Gamma(2.0, 1.0))], 3)"
constraint = "repeat([[bounded_above(9.0)]], 3)"

[uq_param_7]
prior = "Parameterized(MvNormal(3, 2.0))"
constraint = "repeat([no_constraint()], 3)"
``` 

The idea here is that the "repeat(...)" expression for the constraint should be valid Julia code (without the quotation marks of course), such that the corresponding `ParameterDistributions` could be constructed e.g. in a REPL according to:

```
# parameter distribution for uq_param_6 (note that the parameters are generated internally here)
pd6 = ParameterDistribution(
        repeat([Parameterized(Gamma(2.0, 1.0))], 3), 
        repeat([[bounded_above(9.0)]], 3), 
        ["uq_param_6_(1)", "uq_param_6_(2)", "uq_param_6_(3)"])

# parameter distribution for uq_param_7
pd7 = ParameterDistribution(
        Parameterized(MvNormal(3, 2.0)),
        repeat([no_constraint()], 3),
        "uq_param_7")
```
I first wanted to simplify the writing of these expressions in the toml (e.g., by not having the user write too many square brackets :-)), but have come to think that it is much simpler to require the expressions to be written the same way you would write them if you were to actually call the `ParameterDistribution` constructor. This also allows the user to test the validity of the expressions by simply trying to pass them to the `ParameterDistribution` constructor and checking whether that throws an error or not.

test/uq_test_parameters.toml
bielim: I had added the "type" keys in analogy to the format of the "regular" CliMA parameters, but you are right that the information remains unused here, so I deleted all "type" information.

test/uq_test_parameters.toml
bielim: The "type" keys have been deleted (see comment above).

odunbar (approved):
LGTM! Thanks for all the work!

I think we will need to revisit e.g. the exact toml parsing for these broadcasts and alike. But this is an excellent framework you've put together. 

PS One final thing, a warning appears when i run `julia --project ] test` saying the way you call MvNormal is out-of-date. Could you correct this before merging. 
Specifically: `Warning: MvNormal(d::Int, σ::Real) is deprecated, use MvNormal(LinearAlgebra.Diagonal(FillArrays.Fill(σ ^ 2, d))) instead.`

bielim:
Interestingly I don't get the warning about the call to MvNormal, even after upgrading to Julia 1.7.2. I am using `Distributions.jl` v025.53, which is the latest release. I will therefore leave the MvNormal parameter example as is for the time being. If needed, I can add the ability to parse expressions of the sort `MvNormal(LinearAlgebra.Diagonal(FillArrays.Fill(σ^2, d)))` later on (it wouldn't work now as the cost of avoiding the evil `eval` is that the parser doesn't work for any given string representing Julia code but only for a subset of strings representing certain ways of constructing distributions.

PR #70, opened 2022-04-07
add more microphysics params for EDMF calibration

trontrytel:
I want to try more things on the EDMF calibration side. Therefore I need more parameters. 

Will do a patch release after this lands.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CLIMAParameters"
 uuid = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 authors = ["Charles Kawczynski <kawczynski.charles@gmail.com>"]
-version = "0.4.2"
+version = "0.4.3"
 
 [deps]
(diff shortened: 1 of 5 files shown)


charleskawczynski:
Aren't these breaking changes because the name is changing? This is non-breaking if we just add new parameters. If you know for sure that this parameter is not used anywhere, then fine with me to make things move faster.

trontrytel:
bors try

trontrytel:
> Aren't these breaking changes because the name is changing? This is non-breaking if we just add new parameters. If you know for sure that this parameter is not used anywhere, then fine with me to make things move faster.

Oh damn. You are right. I will keep the name for one of them the same so that I can keep it as a patch release.

trontrytel:
bors try

trontrytel:
@charleskawczynski - should be good now?

charleskawczynski (approved):

PR #72, opened 2022-05-09
add gravitational constant and earch mass

yairchn:
I am adding here the universal gravitational constant and the earth's mass. These allow us to compute the Geopotential for deep atmospheres.


charleskawczynski (approved):

trontrytel:
Looks like this wasn't merged and wasn't needed in the last half a year? I'll close it now and re-open when needed.

PR #73, opened 2022-05-10
move non-equilibrium params to a separate module

trontrytel:
This PR
 - moves non-equilibrium parameters to their own category `Microphysics_ne`
 - renames `Microphysics` to `Microphysics_1M`
 
 Hope this helps with packaging patameters into boxes @odunbar @costachris


odunbar (approved):
Am I right in thinking that `τ_cond_evap`, `τ_cond_dep` are only used in the nonequilibrium case (and not in the equlibrium case)? If so, your boxing makes sense and LGTM!

trontrytel:
Can we then package all microphysics parameters into one box? @odunbar 

It's confusing for me to have them structured differently in different places. I would rather have them in one group. Did you ever actually test that splitting them up saves some performance?

odunbar:
@trontrytel 
The fact is, changes to CP as you have done them has apparently (Charlie probs knows better) further changes all the way upstream. Soon this code will be removed, so it is temporary (I am however fine with it, if it is required, or if it turns out there isn't additional work for you guys, i was under the impression that this would be a small change)

RE new CP interface. The splitting into boxes are not really to do with performance. The 1-box solution, would require model runs to define *all* parameters for *all* microphysics schemes (locally or in default). I presume that people running a 0-moment scheme don't want to have to define 1-moment parameters and vice versa. My advice at this stage: If you ever plan adding further schemes I would advise boxes now, but if the 1-moment is the final scheme to be developed, then one box could work.

PR #79, opened 2022-05-20
Remove old method support

charleskawczynski:
This PR:
 - Removes support for the old method-based approach to handling clima parameters
 - Updates the tests accordingly,


src/file_parsing.jl
```diff
@@ -60,6 +46,18 @@ struct AliasParamDict{FT} <: AbstractParamDict{FT}
     data::Dict
     "either a nothing, or a dictionary representing an override parameter TOML file"
     override_dict::Union{Nothing, Dict}
+    "Alias->name map"
+    alias_to_name_map::Dict
+end
+
+function AliasParamDict{FT}(
+    data::Dict,
```
ilopezgp, line 54: are the data entries indexed by names or aliases, could we specify this in a constructor docstring? And I assume the override_dict has the same structure?

ilopezgp (commented):
I am confused, where do the new parameter defaults live? I thought there would be a default TOML in this package containing such information.

ilopezgp:
> I am confused, where do the new parameter defaults live? I thought there would be a default TOML in this package containing such information.

Never mind, I just saw the default toml was added in a previous PR.

src/file_parsing.jl
charleskawczynski: This is just an outer constructor that determines the `alias_to_name_map` (perhaps it should be an inner constructor, though). I believe that the data entries are always indexed by the names (correct me if I'm wrong, @odunbar).

Maybe @odunbar can suggest a better doc string? I believe, or would assume, that the override dict is restricted to a subset of the input dict?

charleskawczynski: @ilopezgp, since this PR is not changing the existing interface or user-facing behavior, let's open this as an issue if it's about the name of the type.

charleskawczynski: I'm actually planning on splitting this part out since it's not really related to removing old method support.

charleskawczynski:
I'm beginning to think that this should be merged sooner than later. Also, while it's a breaking change, every package depending on CP.jl will be able to release patch releases since all packages are already using the new interface, so it's a bit less of a headache.

odunbar:
@charleskawczynski  @ilopezgp  RE EDMF, 

we only require defaults here, parameters are not required to have defaults, however if you feel they will have to be almost always set to the same value across different experiments etc. the it would be sensible to add them. 

I'd suggest a new PR to add some useful subset (or perhaps all) of the EDMF list into defaults.toml from the "old ClimaParameters".

odunbar:
I've just been now informed that we are no longer using Atmos.EDMF or Atmos.Subgridscale parameters (which are now arising from the namelists). These can be added if necessary into defaults.toml in a later PR, as we deprecate the namelists

PR #83, opened 2022-05-24
Improve names struct -> dict

charleskawczynski:
Parameter struct I think is a bit less informative than parameter dict, since it's really just a wrapper around a dict. This PR changes the name to use `dict` over `struct`.

It's probably better that we do this now before spreading to other repos.

I think this will also help differentiate the param dict vs the param "box", which is also a struct but does not necessarily behave like a dict.


odunbar:
how about `toml_dict`? or `param_file_dict`? As it doesnt just contain values, it contains all information from the parameter file

odunbar (approved):
LGTM!

PR #32, opened 2020-06-02
have integers where possible

trontrytel:

PR #87, opened 2022-06-01
Add placeholder for spline functions for orbital parameters

claresinger:
For now these are still set by Insolation.jl. See [PR #32](https://github.com/CliMA/Insolation.jl/pull/32)


src/parameters.toml
```diff
@@ -538,11 +538,23 @@ alias = "eccentricity_epoch"
 value = 0.016708634
 type = "float"
 
-[longitude_perihelion]
-alias = "lon_perihelion"
-value = 4.938188299449
+[orbit_obliquity_spline]
```
charleskawczynski, line 541: I think changes in this file are the only ones really needed since we're moving to the new interface relatively soon

claresinger:
Is there a way to point Insolation.jl to this particular branch of CLIMAParameters? Locally this works fine to add this branch, but it's failing to install the dependencies here: https://github.com/CliMA/Insolation.jl/pull/32.

Or, could we just merge this real quick as a placeholder?

PR #88, opened 2022-06-03
updates for Insolation.jl

claresinger:
- remove `lon_perihelion` parameter
- update `epoch` to be given as a String

@charleskawczynski Do you want to wait until you merge the changes to the code structure? Or can we merge this sooner?


charleskawczynski:
Does any other package use `lon_perihelion`? Deleting it is a breaking change, and upgrading to breaking changes of CLIMAParameters is a lot of work. Can we just not delete it for now?

src/Planet/planet_parameters.jl
```diff
@@ -55,12 +55,11 @@ Planet.grav(ps::AbstractEarthParameterSet) = 9.81
 Planet.year_anom(ps::AbstractEarthParameterSet) = 365.26 * Planet.day(ps)
 Planet.orbit_semimaj(ps::AbstractEarthParameterSet) = 1 * astro_unit()
 Planet.tot_solar_irrad(ps::AbstractEarthParameterSet) = 1362.0
-Planet.epoch(ps::AbstractEarthParameterSet) = 2451545.0 * Planet.day(ps)
+Planet.epoch(ps::AbstractEarthParameterSet) = DateTime(2000,01,01,11,58,56,816)
```
charleskawczynski, line 58: This is also a breaking change..

src/parameters.toml
```diff
@@ -511,9 +511,9 @@ type = "float"
 
 [epoch_time]
 alias = "epoch"
-value = 211813488000
-type = "float"
-description = "derived: 2451545.0 * [day]"
+value = DateTime(2000,01,01,11,58,56,816)
+type = "DateTime"
+description = "DateTime(2000,01,01,11,58,56,816)"
```
charleskawczynski, line 516: @odunbar, will this work?

claresinger:
> 

I can't imagine any other package uses it.

claresinger:
> Does any other package use `lon_perihelion`? Deleting it is a breaking change, and upgrading to breaking changes of CLIMAParameters is a lot of work. Can we just not delete it for now?

I can't imagine any other package uses it. But if it's easier to have unused parameters floating around then we can just keep it.

src/Planet/planet_parameters.jl
claresinger: @tapios This is what you had in mind, right?

tapios: Yes, something like this. I suspect you cannot have functions in the new ClimaParameters. So it may be preferable to have epoch as a string parameter, with value "2000-01-01T11:58:56", which then is parsed to Julian date inside insolation.

src/parameters.toml
```diff
@@ -511,9 +511,9 @@ type = "float"
 
 [epoch_time]
 alias = "epoch"
-value = 211813488000
-type = "float"
-description = "derived: 2451545.0 * [day]"
+value = "2000-01-01T11:58:56.816"
+type = "String"
+description = "J2000 epoch (Jan 1, 2000 11:58:56.816 UTC) as string"
```
claresinger, line 516: @odunbar @charleskawczynski Is this compatible with the new format?

odunbar: Right now we have functionality to allow `AbstractFloat` and `AbstractVector{AbstractFloat}`.
Using the type parameter we could add functionality to add reading of `string` type. Would require a PR.

claresinger: If I stored it as a vector `[2000, 01, 01, 11, 58, 56, 816]` - that would work?

claresinger:
@odunbar I think we need to update the tests? The line [`k_value ≈ cp_k(param_set_cpp)`](https://github.com/CliMA/CLIMAParameters.jl/blob/d3822a0574aaec9eaf56e4db8b2a4c79dafb0205/test/toml_consistency.jl#L109) doesn't work for comparing strings, right?

trontrytel:
What is the status on this one? @claresinger - should we update the new toml files with those parameters?

I think that it's better to delete unused parameters, even if it means a breaking change. Maybe we could collect a couple of such changes and then do it together?

claresinger:
Hi @trontrytel. The status on this was 1) waiting for parameters to accept string values (which has happened) and 2) waiting to delete the unused `lon_perihelion` param because it will be a breaking change and I guess we decided it was easier to just keep it around for now. If you want to collect together a handful of breaking changes, that's great! I'll update the PR now.

claresinger:
bors try

trontrytel:
Sounds good, thank you @claresinger! I also have one name change that would be a breaking thing to merge in. I'll collect the two together and do a breaking release. (And maybe see if other people have some breaking changes they want to make too). @nefrathenrici - should we do it before or after your aerosol PR?

nefrathenrici:
@trontrytel Feel free to do it before my PR, I don't mind adding a minor release afterwards if needed.

szy21:
These changes are never merged - can we merge it now?

tapios:
> These changes are never merged - can we merge it now?

I don't see why not.

szy21:
It looks like we can have strings in parameters now. I'm going to add a new parameter `epoch_time_string` so that it will not break anything.

tapios:
> It looks like we can have strings in parameters now. I'm going to add a new parameter `epoch_time_string` so that it will not break anything.

I would avoid having redundant parameters. 

Also, the newest version of ClimaParams (not yet released), uses `mean_anomaly_at_epoch` instead of `mean_anomalistic_at_epoch`--so we'll need to change that too.

szy21:
> > It looks like we can have strings in parameters now. I'm going to add a new parameter `epoch_time_string` so that it will not break anything.
> 
> I would avoid having redundant parameters.
> 
> Also, the newest version of ClimaParams (not yet released), uses `mean_anomaly_at_epoch` instead of `mean_anomalistic_at_epoch`--so we'll need to change that too.

I just want to avoid breaking Insolation (changing variable name is easy), but I can remove the redundant one. It would be easy enough to change it back.

PR #89, opened 2022-06-06
[WIP] add reading functionality for string and int

odunbar:
## Purpose
To add functionality in `get_parameter_values!` to return more types

## In this PR
- types now interpreted from toml file `type="..."` with options: `"float"`, `"integer"`, or `"string"`
- `"string"` is the default if toml parameter is untyped
- both integer and float are provided when parameter struct created

## TODO 
- replace integer and float types with keyword option "precision" (as unlikely we would use `Float64` but `Int32`?)
- add explicit unit tests for types


src/file_parsing.jl
```diff
@@ -151,35 +151,68 @@ function log_component!(
...
     for alias in aliases
         for (key, val) in data
             alias ≠ val["alias"] && continue
             param_value = val["value"]
+            param_type = haskey(val, "type") ? val["type"] : "string"
             elem =
-                eltype(param_value) != FT ? map(FT, param_value) : param_value
+                isa(param_value, AbstractVector) ?
+                map(x -> _get_typed_value(pd, x, param_type), param_value) :
+                _get_typed_value(pd, param_value, param_type)
```
charleskawczynski, line 199:
```suggestion
                map(x -> _get_typed_value(pd, x, param_type), param_value)
```
I think we might be able to just do this? can you try it out?

src/file_parsing.jl
```diff
@@ -151,35 +151,68 @@ function log_component!(
...
     data = pd.data
     ret_values = map(names) do name
         param_value = data[name]["value"]
+        param_type =
+            haskey(data[name], "type") ? data[name]["type"] : "string"
         elem =
-            eltype(param_value) != FT ? map(FT, param_value) : param_value
+            isa(param_value, AbstractVector) ?
+            map(x -> _get_typed_value(pd, x, param_type), param_value) :
+            _get_typed_value(pd, param_value, param_type)
```
charleskawczynski, line 215:
```suggestion
            map(x -> _get_typed_value(pd, x, param_type), param_value)
```
Same as above, does this work?

src/file_parsing.jl
odunbar: I don't use `eltype` in `_get_typed_value` so not sure - I'll give it a go

odunbar: Ahh it actually has an issue with `string` - it tries to map it to each individual character

charleskawczynski: Ah, right, the original is fine.

docs/src/API.md
```diff
@@ -27,6 +27,7 @@ log_parameter_information
 ```@docs
 log_component!
 get_values
+_get_typed_value
```
charleskawczynski, line 30:
```suggestion
```
I don't think we need to add this to the API docs since its not user-facing.

src/file_parsing.jl
```diff
@@ -151,35 +151,68 @@ function log_component!(
...
+"""
+    _get_typed_value(pd,val,valtype)
+
+enforces `val` to be of type as specified in the toml file
+- `float_type(pd)` if type=\"float\"
+- `Int` if type=\"integer\"
+- `String` if  type=\"string\"
+Default type of `String` is used if no type is provided.
+"""
+function _get_typed_value(pd, val, valtype)
```
charleskawczynski, line 163: I'm wondering if we should also pass the key (i.e., `name` or `alias`) into this function so that we get a more informative error message if/when this errors. Users may not know which type is wrong only based on the value.

odunbar: Yeah I also thought this - would be happy to do so

docs/src/API.md
odunbar: This breaks the docs because there is a docstring in the `file_parsing.jl` file - is there a way of getting past that?

src/file_parsing.jl
odunbar: Although i would say the error message does tell you which "type" i.e the `type = "incorrect"` statement here. Not the value. (So it would be searchable)

docs/src/API.md
charleskawczynski: Yes, let's add `checkdocs = :exports,` to the kwargs in `docs/make.jl`'s call to `makedocs`. This will allow documenter to only check docs for exported functions, which I think makes sense anyway.

Otherwise, we could change the docstring in that function to a comment, but I think using docs is better since we can then use REPL's help functionality.

odunbar: OK I have removed the "internal functions section and added the flag

docs/make.jl
```diff
@@ -26,6 +26,7 @@ makedocs(
     sitename = "CLIMAParameters.jl",
     format = format,
     clean = true,
+    checkdocs = :exports,
```
claresinger, line 29: This makes sense to me.

claresinger (approved):
Looks good! Thanks @odunbar

charleskawczynski (approved):
LGTM, let's just squash first.

PR #98, opened 2022-07-24
Add more parameters, bump major version

trontrytel:
# PULL REQUEST

## Purpose and Content
I need more parameters in CloudMicrophysics.jl and TurbulenceConvection.jl

Not sure if this is the correct workflow. I thought one of the benefits of the new way of handling parameters is that I would not have to open PRs to repos different than the one that needs the parameters? 

This is also a breaking change because I need to change one parameter name.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CLIMAParameters"
 uuid = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 authors = ["Climate Modeling Alliance"]
-version = "0.6.3"
+version = "0.6.4"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -334,6 +334,11 @@ alias = "E_rai_sno"
 value = 1
 type = "float"
 
+[prescribed_cloud_droplet_number_concentration]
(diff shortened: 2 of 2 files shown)


odunbar (commented):
LGTM, 

RE workflow, If parameters are needed to be called in across multiple repositories and there is a sensible default value you will typically be using, then it makes sense to define this value here. If however you are still prototyping, if parameters don't have sensible defaults, or if they are only within some corner of the code, e.g. needed for a single option of some parameterization, then you can just define them within your local experiment override files (which don't require communication with `ClimaParameters.jl`).

odunbar (approved):

charleskawczynski:
This can be a patch update! (also, `(x.y.z) = (major.minor.patch)`)

trontrytel:
I don't think its a patch update since I changed a name of one of the parameters

charleskawczynski:
> I don't think its a patch update since I changed a name of one of the parameters

It is so much easier to do a patch release, can we just not rename the parameter?

trontrytel:
The new name is the correct one and the old one is misleading. I'll change it back to the misleading one because I know how much work it would be to do a major release here. 

I thought that our decentralised approach to parameters would prevent us from having to make such choices. Seems like this is not the case. The alternative would be to delete those parameters from here and add them locally in TC.jl. But since 3 of them are already defined here that would also be a breaking change...

Issue #99, opened 2022-07-26
Variable name changes


charleskawczynski:
This issue is meant to track variable name changes that we'd like to make. Since changing variable names involves releasing a breaking version, which requires propagating through many packages, it makes sense to apply these changes in batches to avoid excessive work.

Here is a list of names we'd like to change:

 - `microph_scaling ` -> `microph_scaling_evap`

cc @trontrytel

trontrytel:
Thank you @charleskawczynski 

I was thinking the same thing. We should collect a bunch of breaking changes to Parameters and weigh in when the number is big enough to warrant the work to propagate the change throughout all the repos

odunbar:
https://docs.google.com/spreadsheets/d/1T3wXWv0tc6O4YuN-7OxMDtezV61RXcUO6DP-eOCXnUc/edit#gid=0

This is something we have already started here. Please add to the list where necessary

trontrytel:
Sorry for replying late! @odunbar why a separate spread sheet and not a github issue? 

Do you have a lot of the breaking changes parameters piled up there? I was hoping to change the above name and to merge https://github.com/CliMA/CLIMAParameters.jl/pull/88 and do a breaking release soon(ish). Do you have anything else that could be merged within a week that would results in a breaking change?

odunbar:
I guess as I've [the link to this sheet elsewhere](https://docs.google.com/spreadsheets/d/1T3wXWv0tc6O4YuN-7OxMDtezV61RXcUO6DP-eOCXnUc/edit#gid=0) the privacy idea is somewhat out the window.

Last year we asked all the groups to contribute their parameters, units, naming conventions, descriptions to a shared sheet. For sharing and editing it was the most convenient format at the time. I do not think it is something we need to keep up-to-date, but should act more as an archive/or starting point for most parameters that are (likely) already in the ClimaParameters list. I expect as every parameter used to be called like `T` or `kappa`, that breaking changes will be frequent as naming conventions change to the more descriptive style.

Nothing should affect your current change.

sajjadazimi:
I replaced microphysics SB2001 scheme by SB2006 which expands SB2001 by more equations. As a result, the parameters of SB2001 are now redundant. I'd like to remove the following list of parameters in the next breaking release:
 - `kc_SB2001`
 - `kr_SB2001`
 - `xstar_SB2001`
 - `ν_SB2001`
 - `A_phi_au_SB2001`
[...]

PR #107, opened 2023-03-25
Adding parameters of the double-moment microphysics scheme of Seifert and Beheng (2006)

sajjadazimi:
## Purpose 
Adding parameters of the double-moment microphysics scheme of Seifert and Beheng (2006)

<!---
Review checklist

I have:
- followed the codebase contribution guide: https://clima.github.io/ClimateMachine.jl/latest/Contributing/
- followed the style guide: https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/
- followed the documentation policy: https://github.com/CliMA/policies/wiki/Documentation-Policy
- checked that this PR does not duplicate an open PR.

In the Content, I have included 
- relevant unit tests, and integration tests, 
- appropriate docstrings on all functions, structs, and modules, and included relevant documentation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "CLIMAParameters"
 uuid = "6eacf6c3-8458-43b9-ae03-caf5306d3d53"
 authors = ["Climate Modeling Alliance"]
-version = "0.7.2"
+version = "0.7.3"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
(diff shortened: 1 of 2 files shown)


trontrytel:
LGTM! The only comment is that if you are deleting the parameters, then technically it should be a "breaking" release (so `v0.8.0` instead of `v0.7.3`). The reason is that the patch release will be automatically updated across all the repos and CIs, but if the parameter is suddenly not there it could break the CIs.

Doing a breaking release for CliMA Parameters is a huge hassle, because everything depends on them and we would have to update the dependencies everywhere. So I would suggest keeping the SB2001 parameters and just adding the SB2006 parameters. This way we can just do a patch release and nothing should be broken. 

Could I then just ask you to add them to the pileup here: https://github.com/CliMA/CLIMAParameters.jl/issues/99 So that when we finally get enough momentum to do a breaking release we will delete them?

PR #109, opened 2023-05-16
Add function to merge multiple TOML files

nefrathenrici:
## Purpose 

This PR adds a function `merge_toml_files` which will merge multiple TOML files, effectively allowing the user to construct a `toml_dict` from more than 2 TOML files.
This will be useful for using multiple case-specific TOML files in addition to the default TOML file. 
This also does not break any of the existing CP interface.
## To-do

- [ ] Add a function which writes a `toml_dict` to a file.

## Content

- More specifically, `merge_toml_files` reads in an arbitrary number of TOML files and merges them into one dictionary.
Overlapping keys are overridden in the order that they are read in and a warning is printed each time an entry is overwritten.
Then, this dictionary is written to a temporary file and returns the filepath, which can be passed into `create_toml_dict`.
- Tests for this are included as well.
- Cleaned up the imports for the other tests.

----

### DIFF
--- a/src/file_parsing.jl
+++ b/src/file_parsing.jl
@@ -173,8 +173,6 @@ function _get_typed_value(
     end
 end
 
-
-
 """
     get_values(pd::AbstractTOMLDict, names)
 
@@ -252,6 +250,33 @@ get_parameter_values(
     pd::AbstractTOMLDict,
     names::Union{NAMESTYPE, AbstractString},
 ) = get_parameter_values!(pd, names, nothing)
+
+
+"""
+    merge_toml_files(filepaths; override)
+
+Parses and merges all of the given TOML filepaths and returns them as a Dict.
+This allows a toml_dict to be constructed from multiple TOML files.
+By default, non-unique TOML entries are not allowed, but this can be
+changed by setting `override = true`.
+"""
+function merge_toml_files(filepaths; override = false)
+    merged_dict = Dict{String, Any}()
+    for filepath in filepaths
+        toml_data = TOML.parsefile(filepath)
+        for (table_name, table_data) in toml_data
+            if haskey(merged_dict, table_name)
+                override || error("Duplicate TOML entry: $table_name")
+                @warn """
+'$table_name' is being overwritten by '$filepath'
+Current entry: $(merged_dict[table_name]["type"])($(merged_dict[table_name]["value"]))
+New entry: $(table_data["type"])($(table_data["value"]))"""
+            end
+        end
+        merge!(merged_dict, toml_data)
+    end
+    return merged_dict
+end
 
 """
     check_override_parameter_usage(pd::ParamDict, strict)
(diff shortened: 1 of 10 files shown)


Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.7.4"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
```
charleskawczynski, line 8: Let's move this to the `test/` environment, CLIMAParameters doesn't directly depend on `SafeTestsets`.

src/file_parsing.jl
```diff
@@ -253,6 +253,35 @@ get_parameter_values(
...
 
+
+"""
+    merge_toml_files(filepaths...)
+
+Parses and merges all of the given TOML filepaths and stores them in a temporary file.
+This allows a toml_dict to be constructed from multiple TOML files.
+Warns when there are non-unique TOML entries.
+"""
+function merge_toml_files(filepaths...)
```
charleskawczynski, line 264: AI like the idea of adding a merge functionality. There is `Base.merge`, and arguably we should subtype `AbstractTOMLDict` by `AbstractDict` and just overload `Base.merge` and `Base.merge!`. Thoughts? If so I'd probably get rid of the `open(f-> TOML.print(f, merged_toml), output_file, "w")` below. Also, we might be able to do better with initializing the merged toml dict. Why does `merge_toml_files` return a file?

nefrathenrici: It's meant to return a filepath to pass into `create_toml_dict` as an override, rather than initialize a proper `AbstractTOMLDict`. I figured this would be easier because it doesn't change the API. I think @odunbar had some reasons for this choice as well

charleskawczynski: Is there a reason we need to create a merged toml file? I figured that we would read the two toml files in and create a merged TomlDict in memory. Or is this needed by the UQ team?

odunbar: Related to Charlie's point about actually writing a new TOML. I don't mind whether we write a new TOML, or just have a dict. Perhaps we can do both, we always output a dict, and if the user requests it via a flag we can write a combined TOML file.

nefrathenrici: Is there a use case for a normal `Dict`? We don't have any exposed functions which could turn a `Dict` into a `ParamDict`

odunbar: Ahh my mistake, I thought there was already an option to pass 2 dicts into `create_toml_dict`!

Sorry to add to this PRs scope. But I would be happy to see this addressed here. Basically you just need to change this [function](https://github.com/CliMA/CLIMAParameters.jl/blob/1245d78041fd1604c66c009508bfa28d874d6429/src/file_parsing.jl#L386-L402) to accept dicts instead (and remove the TOML.parsefile)

Then add the new function e.g. 
```julia 
function create_toml_dict(
    ::Type{FT};
    override_file::Union{Nothing, String} = nothing,
    default_file::String = joinpath(@__DIR__, "parameters.toml"),
  kwargs...
) where {FT <: AbstractFloat}
default_dict = isnothing(default_file) ? TOML.parsefile("../parameters.toml") : TOML.parsefile(default_file)
overried_dict = isnothing(override_file) ? nothing  : TOML.parsefile(override_file)

return create_toml_dict(::Type{FT}, default = default_dict, override = override_dict; kwargs...)

File: .github/workflows/JuliaFormatter.yml [hunk 3]
```diff
@@ -31,7 +31,7 @@ jobs:
     - uses: julia-actions/setup-julia@latest
       if: steps.filter.outputs.julia_file_change == 'true'
       with:
-        version: 1.5.4
+        version: 1.8.5
```
charleskawczynski, line 34: Can we split the version updates off from this PR?

odunbar (approved):
Please address Charlies remark and the formatting test failure, then I am happy for this to be merged in

src/file_parsing.jl
charleskawczynski: > How about we always output a combined Dict (that can be turned into a "TOML dict"), and if the user requests it via a flag (e.g. an output filepath) we can also write a combined TOML file.

sounds good, but I think we should have them as separate functions, so that each function does a single thing— users can then use the atomic parts as needed.

charleskawczynski:
Yeah, let’s rebase this, and then start with the merge functionality that accepts the toml dict and an ordinary dict. Once we have that, we can add a wrapper that writes the new one to a file

src/file_parsing.jl
charleskawczynski: > Is there a use case for a normal Dict? We don't have any exposed functions which could turn a Dict into a ParamDict

Combining with the parsed args, so that we can run the code with command line options.

PR #111, opened 2023-05-23
Update Julia version for GH actions

nefrathenrici:
Updated to 1.8

### DIFF
--- a/.github/workflows/CodeCov.yml
+++ b/.github/workflows/CodeCov.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Julia
       uses: julia-actions/setup-julia@latest
       with:
-        version: 1.7.0
+        version: 1.8
 
     - name: Test with coverage
       env:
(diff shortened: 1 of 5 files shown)


.github/workflows/CodeCov.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Julia
       uses: julia-actions/setup-julia@latest
       with:
-        version: 1.7.0
+        version: 1.8
```
nefrathenrici, line 22: @charleskawczynski Should this be more specific, e.g. 1.8.5?

charleskawczynski: 1.8 should be fine

charleskawczynski (approved):

charleskawczynski:
Ah, we need to update the bors toml for these checks!

nefrathenrici:
bors try

PR #114, opened 2023-07-20
added businger unstable params

wilsonduan10:
## Purpose 

The parameters in the Businger stability functions are hardcoded in the unstable regime. To fix this, I add them as parameters.

## Content

Added b_m and b_h parameters to the toml dict of default parameters which represent the parameters of the stability functions for momentum and heat under unstable conditions. Their default values are given by 15 and 9 respectively, which are the same as those in Businger et al 1971.

----
- [ x ] I have read and checked the items on the review checklist.


nefrathenrici (approved):
Looks good!

akshaysridhar:
Thanks @nefrathenrici

LenkaNovak (approved):
LGTM

PR #122, opened 2023-08-16
Add held suarez parameters, uniqueness test

nefrathenrici:
Content:
- Adds held suarez parameters for ClimaAtmos
- Adds a test ensuring that aliases in the default parameter.toml are unique.


trontrytel (approved):
Do we want to bump the version and do a release?

PR #145, opened 2023-10-23
add area limiter and mixing length parameters

szy21:
## Purpose 

Adds some parameters that are needed for area fraction limiter in entrainment and detrainment. Also adds Smagorinsky coefficient and minimum mixing lenght.

## Content

----

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -2121,6 +2121,18 @@ value = 1
 type = "float"
 description = "TODO: Remove this. Constant entrainment coefficient used for testing EDMF"
 
+[min_area_limiter_scale]
+alias = "min_area_limiter_scale"
+value = 1e-3
+type = "float"
+description = "Constant coefficient that scales the minimum area limiter term in entrainment. Parameter not described in the literature."
+
+[min_area_limiter_power]
+alias = "min_area_limiter_power"
+value = 10
+type = "float"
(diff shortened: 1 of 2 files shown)


src/parameters.toml
```diff
@@ -2139,6 +2151,18 @@ value = 0.12
...
+value = 0.1
+type = "float"
+description = "Constant coefficient that scales the maximum area limiter term in detrainment. Parameter not described in the literature."
+
+[max_area_limiter_power]
+alias = "max_area_limiter_power"
+value = 10
+type = "float"
+description = "Constant coefficient for the exponent in the maximum area limiter term in detrainment. Parameter not described in the literature."
+
```
szy21, line 2165: I can also not add these two parameters and use the ones we have [here](https://github.com/CliMA/CLIMAParameters.jl/blob/945eb72ea84cf5ad8231bcdd7715e8a57dd11e45/src/parameters.toml#L2250-L2260). Just the description is not accurate. What do you think is better?

trontrytel: Maybe we could:
1. Add new ones with the correct descriptions
2. Add the ones we need to delete to this garbage collecting issue: https://github.com/CliMA/CLIMAParameters.jl/issues/99
3. Promise to actually handle this issue before the next sprint ;) @nefrathenrici , @szy21  - What do you think?

trontrytel (approved):

szy21:
@trontrytel I added two more parameters :)

Issue #144, opened 2023-10-20
Overhaul ClimaParameters


nefrathenrici:
# The Climate Modeling Alliance
### Purpose

ClimaParameters is starting to become difficult to use as we scale up the number of parameters and our needs for the interface change. This will address usability, naming conventions, and overriding parameters as particular pain points.
### Cost/Benefits/Risks

The primary risk stems from breaking changes that require people to update downstream repositories. We want to get the interface changes right the first time so that downstream repos don't have to keep changing how they use ClimaParameters. 

The secondary risk is introducing bugs as we add complexity. This will require extensive testing.
### People and Personnel
- Lead: @nefrathenrici
- Collaborators: @odunbar
- Reviewers:

### Components

**Change naming conventions:** The current format requires each parameter to have a long, descriptive name and a shorter `alias` which gets used in code. These both have to be unique, and it forces different models to use the same names in-code. This creates issues for writing clean code: there are currently four parameters that have some variation of the name `T_min`: `T_min`, `T_min_ref`, `T_min_hs`, `H2SO4_sol_T_min`. Since these are all used in different repositories, they could all be called `T_min` if they didn't have to be unique. 

To fix this, the current functionality of the `alias` field and the `AliasParamDict` type will be removed. This functionality will be replaced by parameter name mappings, which form a bijective map from names in the `parameters.toml` file to the names used by packages in-code. This will allow each package to define their own parameters names, so that they don't have to be unique across packages. We will provide a default map so that packages don't need to provide their own name mapping to use climaparameters. This will be based off of the current proper names and aliases.

**Overriding:** The current override syntax is clunky and overly verbose.
Consider the gravity parameter.
```
[gravitational_acceleration]
alias = "grav"
value = 9.81
type = "float"
```
To override this, a separate file or dictionary must be provided with all of the same fields:
```
[gravitational_acceleration]
alias = "grav"
value = 400
type = "float"
```
This is not only annoying to type out, but has also created silent fail issues when parameters weren't being overridden because the name and alias fields didn't match the original parameters.
The only information needed to override a parameter should be the name and the new value:
`gravitational_acceleration = 400`.  For new parameters, type information is still needed, so the current syntax will be still be supported as well.

[...]

PR #157, opened 2023-12-11
Add basic tagging

nefrathenrici:
Part of SDI #144 
### Content
This PR adds a new field `tags`, which will be a list of searchable tags that can be used to find the given parameter.
As an initial examples, I have added tags for the SurfaceFluxes.jl parameters.

In the API, these are supported by two new functions - 
- `get_tagged_parameter_names`: Returns a list of the parameters with a given tag.
- `get_tagged_parameter_values`: Returns a list of name-value Pairs of the parameters with a given tag.

For string-matching, punctuation and capitalization is removed. This could be changed to use string distance instead - I am not sure what the needs are.

I have also added tests for `get_tagged_parameter_values` and `get_tagged_parameter_names`


trontrytel:
I will be able to tag my favorite parameters!

nefrathenrici:
@trontrytel I am planning on adding tags for most parameters once the basic interface has been agreed upon!

trontrytel:
> @trontrytel I am planning on adding tags for most parameters once the basic interface has been agreed upon!

Yes, this looks really cool. I think it will improve the readability a lot once fully implemented

docs/src/toml.md
```diff
@@ -35,6 +36,19 @@ type = "float"
 description = "Molecular weight dry air (kg/mol)"
 ```
 
+### Properly tagged parameter
```
odunbar, line 39: One should highlight in the comments that the search for tagging is done via this "fuzzy" matching, 

It may be useful to export fuzzy_match so users can test how their tags are interpreted in the current fuzzy_match function
e.g. currently it may not be clear to users that
```julia
fuzzy_match("Grachev", "gR*__aC.h,._e, --v-") # returns true
fuzzy_match("Grachev", "&gR*__aC.h,._e, --v-") # returns false
```

src/parameters.toml
```diff
@@ -95,44 +95,51 @@ alias = "Pr_0_Businger"
 value = 0.74
 type = "float"
 description = "Pr_0 for Businger universal functions. From Businger et al, 1971. DOI: 10.1175/1520-0469(1971)028<0181:FPRITA>2.0.CO;2."
+tags = ["Businger", "SurfaceFluxes", "UniversalFunctions"]
```
odunbar, line 98: If we put tags in the default file, be aware that this starts to set convention. If so, we should explicitly state the convention in the docs.

I can't quite make out the convention here. It seems to be a struct that will call this value (e.g. `Businger`) but also the components that build this struct `Businger`?), is that the idea?

docs/src/toml.md
```diff
@@ -35,6 +36,19 @@ type = "float"
...
+The parameter below has tags of varying specificity, from the package name to the specific parameter set it is intended for.
+
+```TOML
+[prandtl_number_0_grachev]
```
odunbar, line 43: Highlight that tags are optional (this is not a namespace), and so the name must still be globally unique

docs/src/toml.md
```diff
@@ -35,6 +36,19 @@ type = "float"
...
+alias = "Pr_0_Grachev"
+value = 0.98
+type = "float"
+description = "Pr_0 for Grachev universal functions. From Grachev et al, 2007. DOI: 10.1007/s10546-007-9177-6"
+tags = ["Grachev", "SurfaceFluxes", "UniversalFunctions"]
+```
+These can be searched via `get_tagged_parameter_names` and `get_tagged_parameter_values`.
```
odunbar, line 50: Perhaps be more explicit here, 

"If this convention is followed, to obtain the parameters used to build tagged by "Grachev" struct, one could call for example:
```julia 
# given toml_dict parsed from parameters.toml
grachev_params = get_tagged_parameter_values(toml_dict, "Grachev")
```

src/file_parsing.jl
```diff
@@ -452,3 +452,36 @@ end
...
+
+
+"""
+    get_tagged_parameter_names(pd::AbstractTOMLDict, tag::AbstractString)
+
+Returns a list of the parameters with a given tag.
+"""
+function get_tagged_parameter_names(
+    pd::AbstractTOMLDict,
+    tag::AbstractString,
```
odunbar, line 464: Could this be a list of tags?

src/file_parsing.jl
```diff
@@ -452,3 +452,36 @@ end
...
+    return lowercase(strip_chars(s1)) == lowercase(strip_chars(s2))
+end
+
+"""
+    get_tagged_parameter_values(pd::AbstractTOMLDict, tag::AbstractString)
+
+Returns a list of name-value Pairs of the parameters with a given tag.
+"""
+get_tagged_parameter_values(pd::AbstractTOMLDict, tag::AbstractString) =
+    get_parameter_values(pd, get_tagged_parameter_names(pd, tag))
```
odunbar, line 487: could we do this for a list of tags

odunbar (commented):
Thanks for this! 

1. I had a couple of overarching notes. Mainly, we should also add documentation for (i) how to check whether your tags are equal with fuzzy_match, (ii) how to call the get_tagged_... utilities, (iii) what the convention is for tagging default parameters. 

2. I wasn't expecting any tags to be applied in this PR, perhaps leave this to a future one?

src/parameters.toml
glwagner: Maybe start conservatively with just the package it pertains to (`SurfaceFluxes`) in this case.

nefrathenrici: I agree, it's best to start with just the package name for these parameters. We can determine conventions when we add tags for the rest of the defaults.

nefrathenrici:
>     2. I wasn't expecting any tags to be applied in this PR, perhaps leave this to a future one?
@odunbar What do you think of just adding the "SurfaceFluxes" tag to the relevant parameters?

odunbar (commented):
Thanks for the updates! I have one further point.
- Please change `tags` to `tag`. I notice that every other field is singular-named, even if it can take an item or vector-of-items as an value.

If it is convenient to add a tags then one can do this here. My concern is that changing these tags requires a PR (breaking if you remove tags) to this repo so make sure they are useful before they are added.

After this i will be happy for merge

nefrathenrici:
@odunbar I have finally dealt with your last comments! I removed all of the tags from the default parameter file and renamed the `tags` field to `tag`.

docs/src/toml.md
```diff
@@ -12,7 +12,8 @@ A parameter is determined by its unique name. It has possible attributes
 3. `type`
 4. `description`
 5. `prior`
-6. `transformation`
+6. `tags`
```
odunbar, line 15: `tag`

odunbar (approved):
One `tags` slipped through the gap, also there are some functions missing from codecov could you add a quick test for `get_tagged_parameter_names` and `get_tagged_parameter_values` for the vector case?

Then merge away!

PR #160, opened 2024-01-23
Update default value for alpha_rayleigh_uh

szy21:
## Purpose 

We want the default value for `alpha_rayleigh_uh` to be zero. This was missed in ClimaAtmos PR[#2021](https://github.com/CliMA/ClimaAtmos.jl/pull/2021) I think. This PR also bumps patch version.

## Content

----
- [ ] I have read and checked the items on the review checklist.

PR #161, opened 2024-01-25
ice deposition nucleation parameters

amylu00:
added parameters for water based deposition nucleation for kaolinite, feldspar, and ferrihydrate


trontrytel (approved):
LGTM! 

You will also need to bump the patch version of the package. So that we can make a release and use it in CM.jl

See for example: https://github.com/CliMA/CLIMAParameters.jl/pull/160/files

PR #166, opened 2024-02-05
Remove default type, update docs

nefrathenrici:
Content
- Remove the default type and error if a parameter has no type
- Add clearer docs for array parameters
- Update test to match
- Fix the formatter GHA update


odunbar (approved):
LGTM! Glad we are removing the default here

NB seems like the formatter is failing, perhaps this needs to be rerun

nefrathenrici:
@odunbar can we consider #144 closed with this PR? I think that this addresses the remaining points (array params and overrides)

PR #167, opened 2024-02-06
correct doi for abifm desert dust params

amylu00:
## Content

----

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -1047,15 +1047,15 @@ description = "Coefficient, c, for calculating deposition nucleation J_het of Ka
 
 # Microphysics - water activity based immersion freezing/nucleation
 
-[KnopfAlpert2013_J_ABIFM_m_DesertDust]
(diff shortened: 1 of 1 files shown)


amylu00:
> Do you need to change the parameter names? This will need to be a breaking release

Yes, but since it's only used in the gpu/perf/unit tests right now Anna and I were hoping to just do a version bump. Is that no good?

nefrathenrici:
That's fair enough! If they are only used in one place it should be fine.

nefrathenrici (approved):

trontrytel:
But we still need to bump the patch version in the `Project.toml` file. `version = "0.8.5"` -> `version = "0.8.6"`

Those parameters are only used in CloudMicrophysics.jl tests. The packages that use CloudMicrophysics.jl don't use those functions. So I vote to do it as a patch release. We will fix the CM.jl tests quickly and noone will notce

trontrytel (approved):

nefrathenrici:
I will need a patch release for #165 as well so we could bundle the changes together.

PR #168, opened 2024-02-07
fix all the typos

trontrytel:
This is a breaking change so we should bump the version. I can update the CloudMicrophysics and Kinematic1D models when it happens. I think they should be the only ones affected

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -477,304 +477,377 @@ type = "float"
 [precipitation_timescale]
 value = 1000
 type = "float"
+description = "0-moment microphysics precipitation formation timescale [s]"
 
 [specific_humidity_precipitation_threshold]
 value = 5e-6
 type = "float"
+description = "0-moment microphysics precipitation formation threshold as specific humidity [-]"
 
 [supersaturation_precipitation_threshold]
 value = 0.02
 type = "float"
+description = "0-moment microphysics precipitation formation threshold as supersaturation [-]"
 
 [rain_drop_drag_coefficient]
 value = 0.55
 type = "float"
+description = "1-moment microphysics rain drop drag coefficient [-]"
 
 [thermal_conductivity_of_air ]
 value = 0.024
 type = "float"
+description = "thermal conductivity of air [J/m/s/K]"
 
 [diffusivity_of_water_vapor]
 value = 0.0000226
 type = "float"
+description = "diffusivity of water vapor [m^2/s]"
 
 [kinematic_viscosity_of_air]
 value = 0.000016
 type = "float"
+description = "kinematic viscosity of air [m^2/s]"
 
 [condensation_evaporation_timescale]
 value = 10
 type = "float"
+description = "Non-equilibrium microphysics condensation/evaporation time scale [s]"
 
 [sublimation_deposition_timescale]
 value = 10
 type = "float"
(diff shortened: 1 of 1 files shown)


nefrathenrici:
This looks good. Do you think that the `Chen2022_table_B3` should be array parameters instead of separate single values? We could get that change in before the minor release as well.

nefrathenrici (approved):

trontrytel:
> This looks good. Do you think that the `Chen2022_table_B3` set of parameters should be a couple of array parameters instead of separate single values? We could get that change in before the minor release as well.

Yes, that would be nicer. So we should replace for example this:

```
[Chen2022_table_B3_As_coeff_1]
value = 0.00174079
type = "float"
description = "part of AS coefficient for ice terminal velocity parameterization - See Table B3 in Chen et al 2022 DOI: 10.1016/j.atmosres.2022.106171"

[Chen2022_table_B3_As_coeff_2]
value = 0.0378769
type = "float"
description = "part of As coefficient for ice terminal velocity parameterization - See Table B3 in Chen et al 2022 DOI: 10.1016/j.atmosres.2022.106171"

[Chen2022_table_B3_As_coeff_3]
value = 0.263503
type = "float"
description = "part of As coefficient for ice terminal velocity parameterization - See Table B3 in Chen et al 2022 DOI: 10.1016/j.atmosres.2022.106171"
```

with this:

```
[Chen2022_table_B3_As]
value = [0.00174079,  0.0378769, 0.263503]
type = "float"
description = "part of AS coefficient for ice terminal velocity parameterization - See Table B3 in Chen et al 2022 DOI: 10.1016/j.atmosres.2022.106171"
```
?

nefrathenrici:
@trontrytel Yes, that looks perfect!

trontrytel:
ok - I'll change it and test if I can use the elegant way in CloudMicrophysics. When do you want to do the release @nefrathenrici ?

nefrathenrici:
We can do another release whenever is convenient. I think these changes can all be in one PR.
I have also found another typo: `rain_cross_section_size_relation_coefficient_dela`

src/parameters.toml
```diff
@@ -935,155 +1008,60 @@ value = 0.115231
...
-[Chen2022_table_B1_a2_coeff]
-value = -0.263166
-type = "float"
-description = "a2 coefficinet for raindrop terminal velocity parameterization [mm^-bi]. See Table B1 in Chen et al 2022 DOI: 10.1016/j.atmosres.2022.106171"
-
-[Chen2022_table_B1_a3_coeff]
-value = 4.7178
-type = "float"
-description = "a3 coefficinet for raindrop terminal velocity parameterization [mm^-bi]. See Table B1 in Chen et al 2022 DOI: 10.1016/j.atmosres.2022.106171"
+description = "ai coefficinets for raindrop terminal velocity parameterization [mm^-bi]. See Table B1 in Chen et al 2022 DOI: 10.1016/j.atmosres.2022.106171"
```
nefrathenrici, line 1014: There is a typo `coefficinets` -> `coefficients` in many descriptions. It would be good to fix those quickly as well!

nefrathenrici (approved):
Thank you for adding the descriptions!

src/parameters.toml
trontrytel: As long as I add less typos than I remove, things should continue to improve ;) Thank you for caching that. Should be fixed now.

PR #171, opened 2024-02-10
Add T_init_min for thermo sat adjust

charleskawczynski:
This PR
 - adds a `T_init_min` parameter for bracketing temperature for numerical methods in thermodynamics
 - Changes `T_min` to 1 Kelvin, so that we can reach lower temperatures than 150 K

There was discussion of using `T_min = 0.42` 😂

Can we do a patch release after this? or does it need to be breaking? cc @trontrytel, @nefrathenrici

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -1641,8 +1641,12 @@ type = "float"
 value = 273.15
 type = "float"
 
-[temperature_saturation_adjustment_min]
+[temperature_saturation_adjustment_init_min]
 value = 150
+type = "float"
+
+[temperature_saturation_adjustment_min]
+value = 1
 type = "float"
 
 [temperature_saturation_adjustment_max]
--- a/test/test_map.jl
+++ b/test/test_map.jl
@@ -79,6 +79,7 @@ import Thermodynamics.Parameters.ThermodynamicsParameters
         :isobaric_specific_heat_liquid => :cp_l,
         :latent_heat_vaporization_at_reference => :LH_v0,
         :temperature_saturation_adjustment_min => :T_min,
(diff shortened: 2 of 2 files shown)


src/parameters.toml
```diff
@@ -1641,10 +1641,14 @@ type = "float"
 value = 273.15
 type = "float"
 
-[temperature_saturation_adjustment_min]
+[temperature_saturation_adjustment_init_min]
 value = 150
 type = "float"
 
+[temperature_saturation_adjustment_min]
+value = 1
```
trontrytel, line 1649: I vote for 0.42 ;)

test/test_map.jl
```diff
@@ -79,6 +79,7 @@ import Thermodynamics.Parameters.ThermodynamicsParameters
         :isobaric_specific_heat_liquid => :cp_l,
         :latent_heat_vaporization_at_reference => :LH_v0,
         :temperature_saturation_adjustment_min => :T_min,
+        # :temperature_saturation_adjustment_init_min => :T_init_min, # will need updated
```
trontrytel, line 82: Should we delete that one?

trontrytel (approved):

test/test_map.jl
charleskawczynski: I figured I'd leave it for when we (eventually) update the compat. We can't right now, because Thermodynamics doesn't yet use this variable. Having TD itself depend on CP and the CP test env depend on TD makes updating dependencies a bit awkward.

src/parameters.toml
charleskawczynski: I'm fine with either one, but I wouldn't be surprised if we're scratching our heads down the road either way :)

PR #174, opened 2024-02-12
Add Frostenberg et al 2023 parameters

trontrytel:
Maybe we can also merge this before the release? @nefrathenrici


nefrathenrici (approved):
Looks good! I'm going to try and figure out why the `dev` docs aren't updating properly before the new release.

PR #178, opened 2024-02-16
P3 ice nucleation parameters

amylu00:
## Content

----

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -1179,6 +1179,28 @@ value = -10.66873
 type = "float"
 description = "Coefficient, c, for calculating immersion freezing J_het of Illite [-]. From Knopf and Alpert 2013. DOI: 10.1039/C3FD00035D."
 
+# Microphysics - P3 ice nucleation paramerizations
+
+[Thompson2004_c1_Cooper]
+value = 0.005
+type = "float"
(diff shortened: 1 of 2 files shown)


src/parameters.toml
```diff
@@ -1179,6 +1179,23 @@ value = -10.66873
 type = "float"
 description = "Coefficient, c, for calculating immersion freezing J_het of Illite [-]. From Knopf and Alpert 2013. DOI: 10.1039/C3FD00035D."
 
+# Microphysics - P3 ice nucleation paramerizations
+
```
trontrytel, line 1183: Could you also add those two numbers: `0.005`, `0.304`. It's from `P3_deposition_N_i`

src/parameters.toml
```diff
@@ -1179,6 +1179,23 @@ value = -10.66873
...
+[Thompson2004_T_threshold_Cooper]
+value = 233
```
trontrytel, line 1185: We actually already have this temperature also. It's here: https://github.com/CliMA/CLIMAParameters.jl/blob/1f3cf67c4296a2f780e55e85660c6e8cd0395c72/src/parameters.toml#L1651

trontrytel (approved):
I vote to add the two parameters from deposition, but reuse the icenucleation temperature we already have.

Thank you for adding them!

PR #179, opened 2024-02-17
Add SoilCO2Model Land parameters

nefrathenrici:
Adds the following parameters:
D_ref,  D_liq, α_sx, Ea_sx, kM_sx, kM_o2, O2_a, D_oa, p_sx 

These could use some better sourcing, but it may not be necessary.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaParams"
 uuid = "5c42b081-d73a-476f-9059-fd94b934656c"
 authors = ["Climate Modeling Alliance"]
-version = "0.10.1"
+version = "0.10.2"
 
 [deps]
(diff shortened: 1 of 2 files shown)


kmdeck (commented):
I think that `b` and `thetaa_100` are soil type dependent and should be read in from maps.
b is the    "Absolute value of the slope of the line relating log(ψ) versus log(θ) (unitless)", but psi(theta) is the matric potential function which for sure depends on soil type/is not universal.
\theta a_100 is the "Air-filled porosity at soil water potential of -100 cm H₂O (~ 10 Pa)" which also depends on psi(theta) and should depend on soil type.

It would also be nice to add references for the default values that are used @AlexisRenchon what source do you use?

PR #182, opened 2024-02-21
Rename CLIMAParameters -> ClimaParameters

nefrathenrici:
This PR renames all instances of `CLIMAParameters` to `ClimaParameters`
I am follwing this [guide](https://github.com/JuliaRegistries/General?tab=readme-ov-file#how-do-i-rename-an-existing-registered-package) to rename and re-register the package.

Additional changes
- New minor version
- New UUID

Still todo after this PR gets merged
- Register the new package
- Change the repo URL and uppdate the URL in the [registry](https://github.com/JuliaRegistries/General/blob/master/C/CLIMAParameters/Package.toml) for the old repo. This should happen ASAP after this PR is merged.


src/CLIMAParameters.jl
```diff
@@ -1,4 +1,4 @@
-module CLIMAParameters
```
charleskawczynski, line 1: We'll need to rename this filename, too. Also, I recommend that that is done in a separate commit, since git can sometimes have a hard time with grouping filename changes with code changes.

charleskawczynski (approved):
42 seems like just the right number of changes, don't you think @trontrytel?

trontrytel:
> 42 seems like just the right number of changes, don't you think @trontrytel?

Destiny has called upon us!

goerz:
Please note that package names are case-preserving, but not case-sensitive, as far as the registry is concerned. You cannot have two packages in the registry that only differ by capitalization, or rename a package like you are trying to do here. You are stuck with the name `CLIMAParamaters`.

PR #183, opened 2024-02-22
Add bucket model parameters

nefrathenrici:


kmdeck (approved):
We just added a couple more parameters to the Bucket model, which would be good to add to CP as well:
f_bucket = 0.75
f_snow = 0.0
p = 1

https://github.com/CliMA/ClimaLand.jl/blob/501829de0e4aee56408cdca64cbc3a237c5fa38e/src/standalone/Bucket/Bucket.jl#L251

I also double checked another model and I think these would be better defaults:
bucket capacity = 0.2 (m)
critical snow amount = 0.05 (m)

Thank you, Nat!! this is a huge help to us.

kmdeck:
oh, and, we can cite the SLIM model (Laguë et al 2019) for all values

nefrathenrici:
> I also double checked another model and I think these would be better defaults:
bucket capacity = 0.2 (m)
critical snow amount = 0.05 (m)

@kmdeck Thank you for the comments! One question, does critical snow amount refer to the parameter critical_snow_water_equivalent or another one?

kmdeck:
> > I also double checked another model and I think these would be better defaults:
> > bucket capacity = 0.2 (m)
> > critical snow amount = 0.05 (m)
> 
> @kmdeck Thank you for the comments! One question, does critical snow amount refer to the parameter critical_snow_water_equivalent or another one?

critical snow water equivalent, thank you!

PR #191, opened 2024-05-08
Add distributions parameters 2-moment scheme

crocicc:
## Content

----

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaParams"
 uuid = "5c42b081-d73a-476f-9059-fd94b934656c"
 authors = ["Climate Modeling Alliance"]
-version = "0.10.5"
+version = "0.10.6"
 
 [deps]
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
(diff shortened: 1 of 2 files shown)


crocicc:
I saw that in ClimaParams I cannot define parameters with a negative value. Since \mu for the rain distribution is negative, I defined it here as positive, and then I'll change the signs in the documentation and in the 2-moment computations. Is that okay?

trontrytel:
> I saw that in ClimaParams I cannot define parameters with a negative value. Since \mu for the rain distribution is negative, I defined it here as positive, and then I'll change the signs in the documentation and in the 2-moment computations. Is that okay?

Why do you think you can't have negative numbers? I think we do have some negative parameters...

crocicc:
Ok perfect, you're right! Now it's negative!

trontrytel (approved):
LGTM!

You will also need to change the version in the `Project.toml` https://github.com/CliMA/ClimaParams.jl/blob/d19d04343c1113575e09b1b50ee48cf355e25f70/Project.toml#L4

to `0.10.6` so that we can make a patch release.

PR #203, opened 2024-08-14
Add turbulent entrainment parameter vector for EDMF

costachris:
Add turbulent entrainment parameter vector for EDMF + bump patch version.


szy21 (approved):
Do you want to bump the version so you can use it in ClimaAtmos?

PR #205, opened 2024-08-19
ClimaLand: all CO2 fluxes in mol CO2 m-2 s-1

AlexisRenchon:
This PR goes along with [ClimaLand PR #731](https://github.com/CliMA/ClimaLand.jl/pull/731),
we are refactoring ClimaLand to express all CO2 fluxes in mol CO2 m-2 s-1, previously there was some in
g C m-2 s-1 and some in mol CO2 m-2 s-1

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -120,6 +120,11 @@ description = "Constant describing cost of maintaining electron transport (unitl
 
 # SoilCO2Model
 
+[kg_C_to_mol_CO2_factor]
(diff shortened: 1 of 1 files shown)


Sbozzolo:
Note that with change you technically triggered a breaking release of ClimaParams (this can be reverted if you add back the parameter you rename).

Could you open an issue in this repo regarding the lack of documentation and guidance on how to add/remove/rename parameters?

PR #210, opened 2024-10-11
adds scm relaxation timescale parameters

Julians42:
## Purpose 

Adds 8 parameters to allow SCM to relax to forcing file profile towards forcing file profile (for use in GCM-driven cases). 

## Content

----

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -2693,6 +2693,46 @@ value = 1
 type = "integer"
 description = "Number of edmf updrafts."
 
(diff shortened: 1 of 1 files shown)


szy21 (approved):
Just a suggestion. For new users `end_height` may sound like there is no relaxation above that height. Maybe something like `maximum_relaxation_height` is more clear.

akshaysridhar (approved):

costachris (approved):

PR #216, opened 2025-02-28
Add ABIFM and ABDINM parameters for various dust

amylu00:


amylu00:
@trontrytel Could you also bump the patch if it looks good? I don't have the permissions required to do that. Thanks!

trontrytel (approved):
LGTM!  @amylu00 - I think you can bump the version in the Project.toml file by yourself. And I'll make a release of the package once this is merged.

Lets remember to update the values we are adding here based on your calibrations once we have them.

And also lets see if we can have one dust category that matches all the experiments. That would be very useful in the context of running this parameterization in Atmos model at some point

PR #223, opened 2025-03-14
add diffusion scaling factor parameters

szy21:
## Content

----


trontrytel (approved):
Could we also add info on the units?

szy21:
Sounds good. I'm testing it in climaatmos now. I will add units and the news once it works.

PR #230, opened 2025-05-14
Update the maximum Prandtl number value

trontrytel:


src/parameters.toml
akshaysridhar, line 1: https://buildkite.com/clima/climacoupler-coarse-nightly-amip/builds/354 This is the relevant build right ?

akshaysridhar (approved):

src/parameters.toml
trontrytel: Yes

trontrytel: From my understanding, it was most likely breaking in the stable cases. And we tend to get more of those at the beginning because of our initial condition (?) So hopefully past week 19 is a good sign.

PR #231, opened 2025-05-20
Add parameters for detrainment ramp and ustar^3 surface tke flux

costachris:
Add parameters for detrainment ramp and ustar^3 surface tke flux formulation.


trontrytel:
Do you also want to increase the version, make a release, etc so that its available in Atmos?

trontrytel (approved):

trontrytel:
See for example here: https://github.com/CliMA/ClimaParams.jl/pull/230/files

costachris:
> Do you also want to increase the version, make a release, etc so that its available in Atmos?

yes, I am confirming the surface flux coefficient is stable in global runs then I will update here and release.

PR #233, opened 2025-06-23
add MERRA2 prescribed aerosol params

sajjadazimi:
## Purpose 
add MERRA2 prescribed aerosol params

----

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -1801,6 +1801,31 @@ value = 1.12
 type = "float"
 description = "Hygroscopicity parameter kappa for sea salt aerosol [-], see: doi.org/10.5194/acp-7-1961-2007"
 
+[MERRA2_seasalt_aerosol_bin01_radius]
+value = 0.079e-6
+type = "float"
+description = "Dry particle radius associated with sea salt aerosol in bin 01 of the MERRA-2 dataset [m], see: doi.org/10.5067/LTVB4GPCOTK2"
+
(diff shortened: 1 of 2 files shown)


src/parameters.toml
```diff
@@ -1864,17 +1894,17 @@ description = "Power of (S_m^2 / (zeta + 3 * eta)) for an empirically determined
...
 
 [mam3_stdev_accum]
-value = 1.8e-6
+value = 1.8
 type = "float"
 description = "Geometric standard deviation for the accumulation modal distribution of MAM3. Unitless. From Liu et al., 2012. DOI: 10.5194/gmd-5-709-2012"
 
 [mam3_stdev_ait]
-value = 1.6e-6
+value = 1.6
```
ray-chew, line 1907: Nice catch, I checked this as the changes are large:

image

ray-chew (approved):

sajjadazimi:
> Nice catch with the `stdev` values from `mam3`. Are they used anywhere in Atmos model? I wonder if merging it in without testing will not break something in case they are

It seems they are used neither in `ClimaAtmos` nor in `CloudMicrophysics`.

src/parameters.toml
sajjadazimi: I guess this mistake comes from the fact that, unlike in a lognormal distribution, the standard deviation in a normal distribution has the same unit as the data. But here, the values represent the standard deviation of a lognormal distribution, and are hence unitless.

PR #237, opened 2025-07-31
Update specific heat ice

tapios:
## Purpose 
Updates the specific heat of ice, to get a slightly more accurate approximation of saturation vapor pressure over ice (inspired by results in Ambaum 2020, 10.1002/qj.3899)

Review checklist

I have:
- followed the codebase contribution guide: https://clima.github.io/ClimateMachine.jl/latest/Contributing/
- followed the style guide: https://clima.github.io/ClimateMachine.jl/latest/DevDocs/CodeStyle/
- followed the documentation policy: https://github.com/CliMA/policies/wiki/Documentation-Policy
- checked that this PR does not duplicate an open PR.

- [x] I have read and checked the items on the review checklist.

### DIFF
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -2160,7 +2160,7 @@ value = 4181
 type = "float"
 
 [isobaric_specific_heat_ice]
-value = 2100
+value = 2070.0
 type = "float"
 
 [temperature_water_freeze]


nefrathenrici (approved):
Thank you! Could you make a minor release? I could do it in a separate PR as well.

nefrathenrici:
> @nefrathenrici - this may affect some of our mse numbers in Atmos. But I don't expect it to break anything. Should we make it a breaking release or not, and just quickly update the reference numbers for Atmos?

I think we should make a breaking (minor) release to make sure we can update the ref counter and the compat at the same time. 

@szy21  made a good point about releasing ClimaParams v1.0.0: every future breaking change after that will require a new major version. This will be inconvenient if we expect frequent breaking changes in the future.

tapios:
> > @nefrathenrici - this may affect some of our mse numbers in Atmos. But I don't expect it to break anything. Should we make it a breaking release or not, and just quickly update the reference numbers for Atmos?
> 
> I think we should make a breaking (minor) release to make sure we can update the ref counter and the compat at the same time.
> 
> @szy21 made a good point about releasing ClimaParams v1.0.0: every future breaking change after that will require a new major version. This will be inconvenient if we expect frequent breaking changes in the future.

It's important here to recognize that a minor change in a parameter, such as here, should not be considered a breaking change. I know it affects MSE tests, but that is a problem with the tests, not with the change. This should be considered a minor modification of model parameters that does not break things once CI etc. are considered properly.

PR #248, opened 2025-08-25
Convert "epoch_time" into DateTime

imreddyTeja:
## Purpose 
PR #246 changed epoch_time from a float to a string. This causes a GPU error in ClimaAtmos:

```
insolation_params is of type Insolation.Parameters.InsolationParameters{Float32, String} which is not isbits.
```

## Content

Previously "epoch_time" was passed in as a string. This creates problem for gpu usage, because strings are not isbitstypes. This commit adds "datetime" as a valtype.
It is also changes "epoch_time" to a datetime in the toml file, which natively supports DateTimes. Users can still pass the "epoch_time" as a string. which is converted to a DateTime.

Alternatively, we can keep epoch_time as a string in ClimaParams, and then convert it into a DateTime in Insolation.jl.

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,12 @@ ClimaParams.jl Release Notes
 ========================
 main
 --------
+
+v0.12.1
+--------
+
 - Log parameter when getindex is called ([#243](https://github.com/CliMA/ClimaParams.jl/pull/243))
+- Change epoch_time to a DateTime ([#248](https://github.com/CliMA/ClimaParams.jl/pull/248))
 
 
 v0.12.0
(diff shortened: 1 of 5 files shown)


szy21:
Thank you! Let me try this with Insolation and ClimaAtmos

szy21:
It works: https://buildkite.com/clima/climaatmos-ci/builds/25807#_ Thanks! Could we make a patch release? (I know, it should probably be a minor release...but it will be faster and I can update the packages in ClimaAtmos today.)

szy21 (approved):

imreddyTeja:
I can make a release, but I'm not sure how important it is to follow Julia's semantic versioning... Maybe @dennisYatunin or @ph-kev have thoughts

szy21 (approved):

nefrathenrici:
Could you update the valid types in the documentation as well? https://github.com/CliMA/ClimaParams.jl/blob/main/docs/src/param_retrieval.md#parameter-types

PR #251, opened 2025-09-18
Add parameters for cloud top height diagnostic

oalcabes:
Adding parameters for cloud top height diagnostic. I just hand tuned these to get somewhat reasonable values for cloud top height -- see below figure. Here I am showing a single column simulation of cli over time, with the cloud top (cltz) calculated in red. The code to actually calculate cloud top height will be in a future ClimaAtmos PR but is quite sensitive to these parameter choices. They can be tweaked in the future to get more accurate values!
Screenshot 2025-09-17 at 5 15 55 PM


oalcabes:
> If it's possible, could you provide some hint on what is the expected range for `cloud_top_k` and `cloud_top_a`. Since they are dimensionless, it's har to guess not knowing the details.
> 
> Also, you will want to bump the patch version to make a release to have them available in Atmos

It's hard to say exactly what they range they should be in on and I think depends on what you want. I just tuned cloud_top_k so that ```w(z) = 1 / (1 + exp(-k * (q_cond - q_threshold) )``` would be close to 0 for values q_c << q_t and close to 1 for q_c ~ q_t. Similarly, I made a small enough that the integral ```\int [w(z) exp(a*z)] dz``` wouldn't blow up, but large enough that it would be close to the cloud top. I have changed each of them individually within about an order of magnitude and if I do that I'm getting ridiculous results. So somewhere within an order of magnitude of these values? Haha. Let me know what you think would be best to say

src/parameters.toml
```diff
@@ -2890,3 +2890,20 @@ description = "Prescribed value of the precipitation fraction (unitless)."
...
+
+[cloud_top_q_threshold]
+value = 1e-8
+type = "float"
+description = "Minimum value for condensate required to calculate top of cloud (kg kg-1)."
+
+[cloud_top_k]
+value = 1e10
+type = "float"
+description = "Parameter controlling steepness of transition between cloud and non-cloud for cloud top calculations (unitless)."
```
tapios, line 2904:
```suggestion
description = "Parameter controlling steepness of (logistic) transition between cloud and non-cloud for cloud top calculations (unitless)."
```

src/parameters.toml
```diff
@@ -2890,3 +2890,20 @@ description = "Prescribed value of the precipitation fraction (unitless)."
...
+
+[cloud_top_a]
+value = 1e-3
+type = "float"
+description = "Parameter controlling strength of favoring cloud top for cloud top calculations (unitless)."
```
tapios, line 2909:
```suggestion
description = "Parameter controlling exponential scale of favoring cloud top for cloud top calculations (unitless)."
```

tapios (approved):
This looks good to me. While I think parameters should generally live in ClimaParams, parameters like these, that are specific to one particular calculation, can also be hardcoded where they are used.

trontrytel:
> > If it's possible, could you provide some hint on what is the expected range for `cloud_top_k` and `cloud_top_a`. Since they are dimensionless, it's har to guess not knowing the details.
> > Also, you will want to bump the patch version to make a release to have them available in Atmos
> 
> It's hard to say exactly what they range they should be in and I think depends on what you want. I just tuned cloud_top_k so that `w(z) = 1 / (1 + exp(-k * (q_cond - q_threshold) )` would be close to 0 for values q_c << q_t and close to 1 for q_c ~ q_t. Similarly, I made cloud_top_a small enough that the integral `\int [w(z) exp(a*z)] dz` wouldn't blow up, but large enough that it would be close to the cloud top. I have changed each of them individually within about an order of magnitude and if I do that I'm getting ridiculous results. So somewhere within an order of magnitude of these values? Haha. Let me know what you think would be best to say

Then lets just merge it here as is and provide some hints in Atmos docs

oalcabes:
> This looks good to me. While I think parameters should generally live in ClimaParams, parameters like these, that are specific to one particular calculation, can also be hardcoded where they are used.

I agree, I think for these in particular given that the results are quite sensitive to these choices I thought it would be nice to add them the standard way so that we can edit them easily when doing things like calibrations.

PR #255, opened 2025-10-01
Do not add getindex for get_parameter_values

### LINKED ISSUE #254: `getindex` in log file
With the change in #243, when logging the parameters, it is a little pointless to place `getindex` on every single line for `used_in`. See below.

```
[thermal_conductivity_of_soil_minerals]
used_in = ["Land", "getindex"]
value = 2.5
type = "float"
description = "Thermal conductivity of soil minerals, excluding quartz (W m⁻¹ K⁻¹). Source: Balland and Arp, J. Environ. Eng. Sci. 4: 549-558 (2005)."

[orbit_eccentricity_at_epoch]
used_in = ["Insolation", "getindex"]
value = 0.016708634
type = "float"
description = "Orbital eccentricity at epoch (unitless)."

[potential_temperature_reference_pressure]
used_in = ["Thermodynamics", "getindex"]
value = 100000.0
type = "float"
description = "Reference pressure for potential temperature calculations ($p_0$) (Pa)."
```

I think `getindex` should appear if the user used it, but `getindex` should not appear if the user use one of the other ways to access the parameters.

ph-kev:
closes #254 - Before this commit, `getindex` was added to "used_in" whenever `get_parameter_values` was called.

### DIFF
--- a/src/ClimaParams.jl
+++ b/src/ClimaParams.jl
@@ -69,6 +69,19 @@ This allows for direct, dictionary-like access to the typed value of a parameter
 
 """
 function Base.getindex(pd::ParamDict, i)
+    val = _getindex(pd, i)
+    log_component!(pd, (i,), "getindex")
+    return val
+end
+
+"""
+    Base.getindex(pd::ParamDict, i)
+
+Retrieves a parameter by its `key`, converting it to the type specified in the TOML file.
+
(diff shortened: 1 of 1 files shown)


nefrathenrici (approved):
Thanks!

PR #296, opened 2026-06-24
New EDMF parameters

sajjadazimi:
## Purpose 
Add EDMF surface condition and cloud fraction parameters

----


szy21 (approved):
It's probably better to make it version 1.1.0 as some parameters names changed, but it doesn't matter much.

costachris (approved):

PR #306, opened 2026-07-28
Sea salt aerosol emission paramters

ziglaser:
## Content

Reference values and coefficients for implementing Gong 2003 and Jaegle 2011 parameterizations for sea salt aerosol emission.

## Notes 
 -  `ssa_size_bin_divisions` is an array of values so that number of bins is derived from bin divisions rather than hard coded separately. IMO best for now but maybe change by aerosol module completion after we have firmer idea of final relationship between bins and optics.
 - `ssa_r_ref`, `ssa_u_ref`, `ssa_T_ref`, `ssa_gong_dim_factor` are SI dimensioned reference factors to non-dimensionalize parameterizations.
 - `ssa_gong_params_A`, `ssa_gong_params_B`, `ssa_gong_params_F`, `ssa_jaegle_params_J` are lists of coefficients rather than many individual entries. This syntax is used for some microphysics parameters in ClimaParams but not super common in the repo. 
 - `ssa_gong_wind_exponent` `ssa_gong_theta`, `ssa_r80_per_dry`, `ssa_residence` are normal parameters

----
- [x] I have read and checked the items on the review checklist.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaParams"
 uuid = "5c42b081-d73a-476f-9059-fd94b934656c"
 authors = ["Climate Modeling Alliance"]
-version = "1.1.5"
+version = "1.1.6"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/src/parameters.toml
+++ b/src/parameters.toml
@@ -3398,3 +3398,58 @@ description = "Steady Beres DC heating weight: Q_t(0)^2 = nogw_beres_steady_dc_f
 value = 1.0e6
 type = "float"
 description = "Largest convective-system scale (m) setting k_min = 2*pi/L in the even-folded horizontal constant H of the steady Beres source (steady/transient ratio depends on it only logarithmically). Source: Beres et al. (2004), DOI: 10.1175/1520-0469(2004)061<0324:AMOICG>2.0.CO;2."
+
+[ssa_residence]
+value = 47520.0
+type = "float"
+description = "Average residence time (sec) of sea salt aerosol mass in AeroCOM3 ensemble. Source: Gliss et al. (2021). DOI:10.5194/acp-21-87-2021"
+
+[ssa_size_bin_divisions]
+value = [0.03e-6, 0.1e-6, 0.5e-6, 1.5e-6, 5.0e-6, 10.0e-6]
+type = "float"
+description = "Double-sided sea salt aerosol dry-radius bin edges (m) from five-bin MERRA-2 scheme, ascending. i.e. bin1 (0.03e-6, 0.1e-6), bin2 (0.1e-6, 0.5e-6)"
+
+[ssa_r_ref]
+value = 1.0e-6
+type = "float"
+description = "Sea salt aerosol reference radius (m) to make Gong parameterization dimensionless."
+
+[ssa_u_ref]
+value = 1.0
+type = "float"
+description = "Wind reference speed (m/s) to make Gong parameterization dimensionless."
+
+[ssa_r80_per_dry]
+value = 2.0
+type = "float"
+description = "Dimensionless ratio of RH80% SSA to dry sea salt radius. Source: Lewis and Schwartz (2004)."
+
+[ssa_gong_wind_exponent]
+value = 3.41
+type = "float"
+description = "Dimensionless exponent on the 10m wind speed in the Gong source function (dimensionless). Source: Gong (2003). DOI: agupubs.onlinelibrary.wiley.com/doi/10.1029/2003GB002079"
+
+[ssa_gong_logfit_mode1]
+value = [0.2157, 0.05545, 17.02]
+type = "float"
+description = "Lognormal mode #1 parameters [flux scale (F) (m⁻² s⁻¹), dry mode radius (r) (μm), lognormal STD (σg) (-)] fit to Gong (2003) spectrum at u_10 = ssa_u_ref."
+
+[ssa_gong_logfit_mode2]
+value = [60.93, 0.0914, 1.813]
+type = "float"
+description = "Lognormal mode #2 parameters [flux scale (F) (m⁻² s⁻¹), dry mode radius (r) (μm), lognormal STD (σg) (-)] fit to Gong (2003) spectrum at u_10 = ssa_u_ref."
+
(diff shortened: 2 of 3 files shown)


src/parameters.toml
```diff
@@ -3357,4 +3357,69 @@ description = "Steady Beres DC heating weight: Q_t(0)^2 = nogw_beres_steady_dc_f
 [nogw_beres_L_system]
 value = 1.0e6
 type = "float"
-description = "Largest convective-system scale (m) setting k_min = 2*pi/L in the even-folded horizontal constant H of the steady Beres source (steady/transient ratio depends on it only logarithmically). Source: Beres et al. (2004), DOI: 10.1175/1520-0469(2004)061<0324:AMOICG>2.0.CO;2."
\ No newline at end of file
+description = "Largest convective-system scale (m) setting k_min = 2*pi/L in the even-folded horizontal constant H of the steady Beres source (steady/transient ratio depends on it only logarithmically). Source: Beres et al. (2004), DOI: 10.1175/1520-0469(2004)061<0324:AMOICG>2.0.CO;2."
+
+[ssa_residence]
+value = 47520
```
akshaysridhar:
```suggestion
value = 47520.0
```

akshaysridhar: (We have a distinct integer parameter type), likewise for the T_ref and u_ref params added

src/parameters.toml
akshaysridhar, line 1: Please also bump the patch `version` in the Project.toml file. 1.1.3 unless a separate patch update PR merges ahead of this PR.

src/parameters.toml
```diff
@@ -3357,4 +3357,69 @@ description = "Steady Beres DC heating weight: Q_t(0)^2 = nogw_beres_steady_dc_f
...
+type = "float"
+description = "Sea salt aersol reference radius (m) to make Gong parameterization dimensionless."
+
+[ssa_u_ref]
+value = 1
+type = "float"
+description = "Wind reference speed (m/s) to make Gong parameterization dimensionless."
+
+[ssa_T_ref]
+value = 1
```
trontrytel: Would it make sense to use a more reasonable value for reference here?

src/parameters.toml
```diff
@@ -3357,4 +3357,69 @@ description = "Steady Beres DC heating weight: Q_t(0)^2 = nogw_beres_steady_dc_f
...
+
+[ssa_r80_per_dry]
+value = 2.0
+type = "float"
+description = "Dimensionless ratio of RH = 80% SSA to dry sea salt radius. Source: Lewis and Schwartz (2004)."
+
+[ssa_gong_params_A]
+value = [4.7, -0.017, -1.44]
+type = "float"
+description = "Dimensionless coefficents in Gong factor A = a[1] * (1 + θ * r̂)^(a[2] * r̂^a[3]). Source: Gong (2003). DOI: agupubs.onlinelibrary.wiley.com/doi/10.1029/2003GB002079"
```
trontrytel: If they are dimensionless, then why do we need to add the reference T, u and r?

src/parameters.toml
ziglaser: The SST dependent parameterization is in Celcius, so this is used by taking (SST [K] - freezing i.e 273.15 [K]) / ref_temp i.e. 1 [K] to non-dimensionalize. I think this is defensible, and picking another reference temp would require recomputing the coefficients.

trontrytel: Yes, that was my point. I think 1K is a strange reference temperature. And I would rather recompute the coefficients in the code

ziglaser: Would it make more sense to you to use freezing as the reference temperature--rescaling the parameters by 273.15^n for each polynomial term, or use 30 K as the reference temperature, since that gives T_hat ~ in [0, 1]?

src/parameters.toml
ziglaser: The full expressions for A and B both have to wind up unitless since we have factors of r^A and 10^{-B^2} in the final parameterization. Units enter in the full parameterization, which has coefficients F, though I pull them out into a dimensioned pre-factor, could have the F coefficients as separate dimensioned constants instead.

src/parameters.toml
ziglaser: Done

ziglaser:
@trontrytel @akshaysridhar Deleted ssa_T_ref, reformatted to floats, bumped to 1.1.3 and added to NEWS.md, ready for re-review

src/parameters.toml
trontrytel: That does sound better. Thank you. Either way works

ziglaser:
@trontrytel Revised for new lognormal Gong approximation, rebased, good for rereview

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "ClimaParams"
 uuid = "5c42b081-d73a-476f-9059-fd94b934656c"
 authors = ["Climate Modeling Alliance"]
-version = "1.1.3"
+version = "1.1.4"
```
trontrytel: I think we are releasing a version now. So you may need to bump it once more

src/parameters.toml
```diff
@@ -3358,3 +3358,58 @@ description = "Steady Beres DC heating weight: Q_t(0)^2 = nogw_beres_steady_dc_f
...
+
+[ssa_residence]
+value = 47520.0
+type = "float"
+description = "Average residence time (sec) of sea salt aerosol mass in AeroCOM3 ensemble. Source: Gliss et al. (2021). DOI:10.5194/acp-21-87-2021"
+
+[ssa_size_bin_divisions]
+value = [0.03e-6, 0.1e-6, 0.5e-6, 1.5e-6, 5.0e-6, 10.0e-6]
+type = "float"
+description = "Sea salt aerosol dry-radius bin edges (m) from five-bin MERRA-2 scheme, ascending."
```
trontrytel: Are those left edges or right edges?

trontrytel (approved):

src/parameters.toml
ziglaser: Both, i.e. bin1 is (0.03e-6, 0.1e-6), bin2 is (0.1e-6, 0.5e-6) ... bin5 (5.0e-6, 10.0e-6). Maybe that's a bad convention, would take suggestions

trontrytel: You can just add that to the description so that it's not ambiguous.

